PR #11, opened 2018-11-29
Various WebGL Perf-related changes

jeffsaremi:
See the commit descriptions for details
New Benchmark numbers need to be reproduced

### DIFF
--- a/lib/backends/webgl/ops/conv.ts
+++ b/lib/backends/webgl/ops/conv.ts
@@ -213,7 +213,8 @@ export class WebGLConv extends Conv {
       inputLayouts: inputs.length === 3 ? [im2colLayout, kLayout, bLayout!] : [im2colLayout, kLayout],
       outputLayout,
       shaderSource,
-      params: {'sharedDim': sharedDim, 'sharedDimReadSize': sharedDimReadSize}
+      params: {'sharedDim': sharedDim, 'sharedDimReadSize': sharedDimReadSize},
+      blockSize: this.calcBlockSize(outputLayout)
     };
   }
   createDotProdRunData(inferenceHandler: WebGLInferenceHandler, programInfo: ProgramInfo, inputs: Tensor[]): RunData {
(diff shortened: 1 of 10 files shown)


hariharans29:
Can we also update the corresponding perf improvements in the main README (the image corresponding to GPU latency)

lib/backends/webgl/ops/conv.ts
```diff
@@ -288,15 +289,15 @@ export class WebGLConv extends Conv {
...
+    const preferredBatchSize = 8;
     if (sharedDim < preferredBatchSize || sharedDim % preferredBatchSize !== 0) {
       return sharedDim;
     }
     return preferredBatchSize;
   }
   protected calcBlockSize(outputLayout: TextureLayout): [number, number]|undefined {
-    const preferredRowCount = 64;
-    const preferredColCount = 64;
+    const preferredRowCount = 32;
```
fs-eire: how do we determine the best block size? does it depend on the GPU spec

lib/backends/webgl/program-manager.ts
```diff
@@ -97,6 +101,7 @@ export class ProgramManager {
         } else {
           this.doDraw(buildArtifact, runData);
         }
+        gl.flush();
```
fs-eire: we can remove the `gl.flush()` call inside `doDraw()`

jeffsaremi:
I will be re-submitting this after fixing a bug that was introduced.

lib/backends/webgl/ops/conv.ts
jeffsaremi: Purely experimental.
These are the best combinations I could come up with on my machine.
In the future, we may decide to expose these to the very savvy users

lib/backends/webgl/program-manager.ts
jeffsaremi: Thank you. that was left by mistake

PR #13, opened 2018-11-30
webgl: various optimizations

jeffsaremi:
See the commit list for details

### DIFF
--- a/lib/backends/webgl/texture-helper.ts
+++ b/lib/backends/webgl/texture-helper.ts
@@ -22,7 +22,6 @@ import {WebGLContext} from './webgl-context';
  */
 export class TextureHelper {
   glContext: WebGLContext;
-  free: Map<string, WebGLTexture[]>;
   gl: WebGLRenderingContext;
   layoutStrategy: TextureLayoutStrategy;
   profiler: Readonly<Profiler>;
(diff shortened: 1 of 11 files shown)


lib/backends/webgl/texture-helper.ts
```diff
@@ -17,47 +17,36 @@ import {WebGLContext} from './webgl-context';
  *    Caching these is crucial to performance. These are In-use Textures
  *   2. textures which are not in use by any current ProgramInfo/Tensor
  *     These are called Free Textures
- * TextureManager is also used to help creating textures. For this it
+ * TextureHelper is also used to help creating textures. For this it
  * uses WebGLContext and TextureLayoutStrategy
  */
-export class TextureManager {
+export class TextureHelper {
```
hariharans29: Nice set of changes, just curious - any reason for this name change - 'TextureManager' -> 'TextureHelper' ?

jeffsaremi: TensorFlow has a class by the same name too. I don't want to think we're copying their stuff

jeffsaremi: and the role of this class has changed. Initially used to be responsible for some elaborate caching of textures. Now it's just a bunch of utilities

Issue #16, opened 2018-12-03
Uncaught (in promise) Error: multiple nodes output to one data value


huangyakun:
When I load a pytorch export onnx model, something wrong happend like this, what's wrong with my operations? I have checked the exported onnx model with onnx.checker. Thanks!
graph.ts:298 Uncaught (in promise) Error: multiple nodes output to one data value: 19
    at t.buildGraph (graph.ts:298)
    at new t (graph.ts:139)
    at Object.from (graph.ts:77)
    at t.load (model.ts:29)
    at session.ts:85
    at t.event (instrument.ts:294)
    at e.initialize (session.ts:81)
    at e.<anonymous> (session.ts:63)
    at onnx.ts:87
    at Object.next (onnx.ts:87)

fs-eire:
thanks for the feedback. could you please share the model so that I can take a look?

huangyakun:
Hi,
     Thanks for the feedback. 
      You can download the onnx from 
      https://drive.google.com/open?id=1kXdkxmyhUpnL_-T5K8-Wb5CyQjTnZc1V
     And another onnx model has the same problem which can be downloaded from
      https://drive.google.com/open?id=1N6eDfXDE-9QmFOOemC9FkjqC_D1NUWk8
Thanks!
Huang

hariharans29:
@fs-eire - For the second issue (operator BatchNormalization reports 'InvalidInputs' because of X.dims.length < 3), the input data X should be atleast 3 dimensional. The first dimension is batch size, the second is channels, and the third onwards are the data dimensions. More details [here](https://github.com/onnx/onnx/blob/rel-1.2.3/docs/Operators.md#inputs-9).

ybrnathan:
Double checked PyTorch BatchNormalization implementation, seems like PyTorch has two versions:
- torch.nn.BatchNorm1d
    Input: (N,C) or (N,C,L)
    Output: (N,C) or (N,C,L)

- torch.nn.BatchNorm2d
    Input: (N,C, H, W)
    Output: (N,C, H, W)

And according ONNX spec, ONNX only support (N, C, H, W) or (N, C, L) and does not support (N, C). In your model, The BatchNormalization node after MatMul input shape is (N, C) which is not supported according to ONNX spec.

hariharans29:
Closing this issue as we root-caused the issue. ONNX spec doesn't support BatchNormalization op taking input of shape (N,C)

PR #20, opened 2018-12-03
graph: fix resolve operator 'Constant'

fs-eire:
This change fixes a bug in graph resolving when there is a 'Constant' operator. This shows in issue #16 ( the first model). @hariharans29 @liuziyue 

===

The problem is, we used same node index for `node` and `nodeProto`, https://github.com/Microsoft/onnxjs/blob/6aa8898b619a3ab888151d96f7489db03ae97a9c/lib/graph.ts#L284-L285

but if there is a 'Constant' node, the index will not match:

https://github.com/Microsoft/onnxjs/blob/6aa8898b619a3ab888151d96f7489db03ae97a9c/lib/graph.ts#L242-L262

### DIFF
--- a/lib/graph.ts
+++ b/lib/graph.ts
@@ -287,9 +287,8 @@ class GraphImpl implements Graph, Graph.Transformer {
           node.outputs.pop();
           node.executeNode = false;
 
-          const tensor = Tensor.fromProto(nodeProto.attribute[0].t);
           this._allData[dataIndex]._from = -1;
-          this._allData[dataIndex].tensor = tensor;
+          this._allData[dataIndex].tensor = Tensor.fromProto(nodeProto.attribute[0].t);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


lib/graph.ts
```diff
@@ -298,6 +274,23 @@ class GraphImpl implements Graph, Graph.Transformer {
...
+          }
+          if (!nodeProto.output || nodeProto.output.length !== 1) {
+            throw new Error(`missing output or incorrect number of outputs for this Constant operator`);
+          }
+          node.outputs.pop();
+          node.executeNode = false;
+
+          const tensor = Tensor.fromProto(nodeProto.attribute[0].t);
+          this._allData[dataIndex]._from = -1;
+          this._allData[dataIndex].tensor = tensor;
```
hariharans29: Line 292 can directly contain Tensor.fromProto(nodeProto.attribute[0].t);. There is no 'if'...'else' like there was previously

PR #25, opened 2018-12-04
Webgl options needed to fix mac/linux issues

jeffsaremi:
See the commits for details


lib/api/onnx.ts
```diff
@@ -14,7 +14,16 @@ export declare namespace Backend {
   /**
    * set options for the WebGL backend
    */
-  interface WebGLOptions {}
+  interface WebGLOptions {
+    /**
+     * set or get the WebGL Context ID (webgl vs webgl2,...)
+     */
+    contextId?: string;
```
fs-eire:
```ts
contextId?: 'webgl'|'webgl2'|'webgl-experimental';
```

lib/api/onnx.ts
```diff
@@ -14,7 +14,16 @@ export declare namespace Backend {
...
+    /**
+     * set or get the flag which determines which format parameter to pass to gl.readPixels()
+     */
+    forceRgbaReads?: boolean;
```
fs-eire: I would prefer to use something like `readPixelsFormat?: 'ALPHA'|'RGB'|'RGBA';`

lib/backends/webgl/base-webgl-context.ts
```diff
@@ -126,12 +126,12 @@ export abstract class BaseWebGLContext implements WebGLContext, Disposable {
   }
   readTexture(
       texture: WebGLTexture, width: number, height: number, dataSize: number, dataType: Encoder.DataType,
-      channels: number): Encoder.DataArrayType {
+      channels: number, forceRgbaReads?: boolean): Encoder.DataArrayType {
```
fs-eire: should we access WebGLOptions from the backend object, instead of passing all those flags from functions to functions

lib/backends/webgl/session-handler.ts
```diff
@@ -39,7 +39,9 @@ export class WebGLSessionHandler implements SessionHandler {
   textureDataCache: Map<Tensor, TextureData>;
   initializers: Set<Tensor>;
 
-  constructor(public readonly backend: WebGLBackend, public readonly context: Session.Context) {
+  constructor(
+      public readonly backend: WebGLBackend, public readonly context: Session.Context,
+      public readonly forceRgbaReads?: boolean) {
```
fs-eire: read flags from backend instead of passing in constructors

lib/api/onnx.ts
jeffsaremi: I am removing forceRGBaReads period. which should resolve this issue as well.

lib/backends/webgl/texture-data-encoder.ts
```diff
@@ -70,13 +70,14 @@ export class RedFloat32DataEncoder implements DataEncoder {
...
   }
   decode(buffer: Encoder.DataArrayType, dataSize: number): Float32Array {
-    if (dataSize < buffer.length) {
-      return buffer.slice(0, dataSize) as Float32Array;
+    if (this.channelSize === 1) {
+      const filteredData = (buffer as Float32Array).filter((value, index) => index % 4 === 0).slice(0, dataSize);
+      return filteredData;
     }
-    return buffer as Float32Array;
+    return buffer.slice(0, dataSize) as Float32Array;
```
hariharans29: Any reason why we need to make a copy here now ?

jeffsaremi: no reason. changed it to subarray() to create another view instead after consult with Yulong

lib/backends/webgl/texture-data-encoder.ts
```diff
@@ -70,13 +70,14 @@ export class RedFloat32DataEncoder implements DataEncoder {
...
   }
   allocate(size: number): Encoder.DataArrayType {
-    return new Float32Array(size * this.channelSize);
+    return new Float32Array(size * 4);
   }
   decode(buffer: Encoder.DataArrayType, dataSize: number): Float32Array {
-    if (dataSize < buffer.length) {
-      return buffer.slice(0, dataSize) as Float32Array;
+    if (this.channelSize === 1) {
+      const filteredData = (buffer as Float32Array).filter((value, index) => index % 4 === 0).slice(0, dataSize);
```
fs-eire: do you want to use .subarray() here as well?

jeffsaremi: why not if it saves us from doing another copy on top of filter()

PR #32, opened 2018-12-05
doc: update operator support table and test ReadMe

hariharans29:
* operator support table is updated to reflect the fact that we support Constant, Dropout, and Identity for all backends (as they are pre-processed out in the graph). This should save the users confusion as to whether they can use they can use their model even though the operator support table doesn't 'Dropout' op for the 'wasm' backend

* ./test/README.md is out of date. Updated to reflect recent changes.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -40,7 +40,7 @@ Replace `test_abs` with any other sub-folder located under the `deps/data/data/t
 
 By default, all available backends will be tested. Use the --backend flag (options: `cpu`, `wasm`, `webgl`) to test specific backends
 ```
-npm test -- model test_abs --backend webgl
+npm test -- model test_abs --backend=webgl
 ```
 
 ### Op Tests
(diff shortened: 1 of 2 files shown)


test/README.md
```diff
@@ -24,32 +24,41 @@ npm test -- unittest
 
 ### Model Tests
 These are tests based on Onnx models and their inputs and outputs are also in binary `ProtoBuf` formats. Data for these
-tests is located under `test/test-data/models` and `test/test-data/nodes`.
+tests is located under `deps/data/data/test/node` and `deps/data/data/test/onnx/v7`.
 
 To run individual Model tests for the included, well-known models (such as Resnet50) type the following:
 ```
-npm test -- model test/test-data/onnx/v7/resnet
+npm test -- model resnet
```
fs-eire: fixit: resnet50

PR #34, opened 2018-12-06
WebGL-related fixes for stability and perf

jeffsaremi:
e667352 (HEAD -> dev/jeff/webgl-perf1, origin/dev/jeff/webgl-perf1) removing redundant call
e8b86eb fixing a bug with Reshape operator and caching texturedata
4fa0804 binding Attributes(vertices) only once
567fa5b Ensuring textures are released at the end of each inference
7d8e7c9 adding --printMatches to verify accuracy visually
6ba453c renaming TextureManager


liuziyue:
Verified that demo works as expected for this one.

PR #35, opened 2018-12-06
tensor: add data size safety check and move validateDims() to utilities

hariharans29:
* Add data size safety check to ensure the data returned by the DataProvider is consistent with the Tensor dims

* Moving validateDims() into utilities (/lib/util.ts) to be leveraged by other modules going forward

### DIFF
--- a/lib/tensor.ts
+++ b/lib/tensor.ts
@@ -126,6 +126,11 @@ export class Tensor {
     // return this.cache!;
   }
 
+  /**
+   * get the number of elements in the tensor
+   */
+  public readonly size: number;
+
   constructor(
       /**
        * get the dimensions of the tensor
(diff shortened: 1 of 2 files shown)


lib/tensor.ts
```diff
@@ -303,25 +299,8 @@ export class Tensor {
...
 
-  for (const n of dims) {
-    if (!Number.isInteger(n)) {
-      throw new TypeError(`Invalid shape: ${n} is not an integer`);
-    }
-    if (n <= 0 || n > 2147483647) {
-      throw new TypeError(`Invalid shape: length ${n} is not allowed`);
-    }
-  }
+  public size: number;
```
fs-eire: readonly

fs-eire: add jsdoc back

lib/tensor.ts
```diff
@@ -122,13 +126,6 @@ export class Tensor {
     // return this.cache!;
   }
 
-  /**
-   * get the number of elements in the tensor
-   */
-  get size(): number {
```
fs-eire: keep property here instead of in the bottom of the class body

PR #38, opened 2018-12-07
tensor: support scalar tensor (and associated changes)

hariharans29:
* Supporting scalar tensor within ONNX.js


fs-eire:
Close this change according to previous discussion. Should be replaced by #40

PR #39, opened 2018-12-07
reshape: cleanup WebGL reshape op and add size safety check in reshape utility

hariharans29:
* Cleanup unused WebGL backend 'reshape' op
* Add size safety check in reshape utility method in /lib/util.ts


lib/backends/webgl/ops/reshape.ts
```diff
@@ -1,45 +1,16 @@
...
-          inputShapes[0].slice(this.dimsToKeep[0]));
-    }
-  }
   run(inferenceHandler: WebGLInferenceHandler, inputs: Tensor[]): Tensor[] {
-    const inputShape: ReadonlyArray<number> = inputs[0].dims.slice();
-    const inputShapes = [inputShape];
-    if (inputs.length > 1) {
-      inputShapes.push(ShapeUtil.calculateReshapedDims(inputShape, inputs[1].integerData));
-    }
+    const reshapedDims = ShapeUtil.calculateReshapedDims(inputs[0].dims, inputs[1].integerData);
```
hariharans29: @fs-eire and @jeffsaremi - While trying to understand the reshape op in webgl - I saw some redundant code (I think!). Are you okay with this cleanup ?

lib/util.ts
```diff
@@ -489,17 +489,23 @@ export class ShapeUtil {
...
+      if (oldTensorSize % newTensorSize !== 0) {
         throw new Error(`the input tensor cannot be reshaped to the requested shape. Input shape: [${
             originalDims}] Output shape: [${shapeHints}]`);
       }
-      reshapedDims[unknownDimension] = originalTensorFlattenedSize / size;
+      reshapedDims[unknownDimension] = oldTensorSize / newTensorSize;
+    }
+    // validate sizes from originalDims and reshapedDims match
+    else {
+      if (newTensorSize !== oldTensorSize) {
```
hariharans29: Adding check here to ensure the sizes of the originalDims and reshapedDims remain the same after reshape.

lib/backends/webgl/ops/reshape.ts
jeffsaremi: this looks good to me. thanks

PR #43, opened 2018-12-11
test: several improvements

fs-eire:
- consume API types instead of internal ones in Model test
- introduce 'bundleMode', with 3 modes: prod, dev and perf
- other trival improvements and fixes

### DIFF
--- a/test/test-runner.ts
+++ b/test/test-runner.ts
@@ -7,12 +7,13 @@ import {onnx} from 'onnx-proto';
 import {extname} from 'path';
 import {inspect, promisify} from 'util';
 
+import * as api from '../lib/api';
+import {fromInternalTensor, toInternalTensor} from '../lib/api/tensor-impl-utils';
 import {Attribute} from '../lib/attribute';
 import {Backend, InferenceHandler, SessionHandler} from '../lib/backend';
 import {TextureData} from '../lib/backends/webgl/texture-data';
 import {Logger, Profiler} from '../lib/instrument';
 import {Operator} from '../lib/operators';
-import {Session} from '../lib/session';
 import {Tensor} from '../lib/tensor';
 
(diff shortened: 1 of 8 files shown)


tools/test-runner-cli.ts
```diff
@@ -608,3 +449,17 @@ function run(config: Test.Config) {
 function saveConfig(config: Test.Config) {
   fs.writeFileSync(path.join(TEST_ROOT, './testdata.js'), `module.exports=${JSON.stringify(config, null, 2)};`);
 }
```
jeffsaremi: I almost created the same thing in my PR. Nice coincidence

jeffsaremi:
Overall I cannot follow the changes. I don't know how all these options make their way from the command line to the test-runner which is where they matter. Nevertheless I trust you have somehow taken care of this. So I'm approving this based on that trust.

test/test-runner.ts
```diff
@@ -92,7 +148,7 @@ async function loadTensorProto(uri: string): Promise<Test.NamedTensor> {
   const buf = await loadFile(uri);
   const tensorProto = onnx.TensorProto.decode(Buffer.from(buf));
   const tensor = Tensor.fromProto(tensorProto);
-  const namedTensor = tensor as NamedTensorImpl;
+  const namedTensor = fromInternalTensor(tensor) as unknown as NamedTensorImpl;
```
liuziyue: Just curious, why is "as unknown" needed here?

test/test-runner.ts
```diff
@@ -28,29 +29,84 @@ const WASM_THRESHOLD_RELATIVE_ERROR = 1.000001;
 /**
  * a simple class that implementes interface Test.NamedTensor
```
liuziyue: nit "implementes" typo here

PR #45, opened 2018-12-11
Tensor Utility API

liuziyue:


lib/api/tensor-transforms.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+// Exposes a set of light-weight tensor(s) transformations
+
+import {Tensor} from './tensor';
+
+export interface UtilityTensorCreators {
+  zeros(shape: ReadonlyArray<number>, dtype: Tensor.NumberOrBoolType): Tensor;
```
hariharans29: Do we need Jsdoc comments for these methods in this file? @fs-eire - what do you think ?

hariharans29:
Will close this and re-open later if needed

PR #47, opened 2018-12-13
wasm: Binary Operators, MatMul operator, shape_utils, and broadcast_utils

hariharans29:
* This change contains some binary operators and MatMul operator for the wasm backend along with the associated shape_utils and broadcast_utils for the wasm backend

* Also contains some very trivial fixes to make the binary-op classes for each of the available backends consistent in their interfaces


lib/backends/wasm/ops/matmul.ts
```diff
@@ -0,0 +1,70 @@
...
+import {WasmBinding} from '../../../wasm-binding';
+import {WasmInferenceHandler} from '../inference-handler';
+
+export class WasmMatMul extends MatMul {
+  run(inferenceHandler: WasmInferenceHandler, inputs: Tensor[]): Tensor[] {
+    const a = inputs[0];
+    const b = inputs[1];
+    let dimsA = a.dims.slice(0);
+    let dimsB = b.dims.slice(0);
+    // If the first argument is 1-D, it is promoted to a matrix by prepending
```
fs-eire: according to the operator's [definition](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html), the preprocess and postprocess should be used by all backends. is it a good idea to put in /lib/ops/matmul.ts

lib/backends/wasm/ops/matmul.ts
```diff
@@ -0,0 +1,70 @@
...
+    const resultData = new Float32Array(outputSize);
+    WasmBinding.getInstance().ccall(
+        '_matmul_f32', [inputs[0].floatData, 'float32ptr'], [inputs[0].dims, 'int32ptr'],
+        [inputs[0].dims.length, 'int32'], [inputs[1].floatData, 'float32ptr'], [inputs[1].dims, 'int32ptr'],
+        [inputs[1].dims.length, 'int32'], [resultData, 'float32ptr', 'out'], [resultData.length, 'int32'],
+        [outputShape, 'int32ptr'], [outputShape.length, 'int32']);
+
+    // Remove prepended dimension if first input is 1d
+    if (a.dims.length === 1) {
+      outputShape = outputShape.slice(0, outputShape.length - 2).concat(outputShape.slice(outputShape.length - 1));
```
fs-eire: is `splice()` a good replacement of `slice()` here?

lib/backends/wasm/ops/matmul.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    // Remove appended dimension if second input is 1d
+    if (b.dims.length === 1) {
+      outputShape = outputShape.slice(0, outputShape.length - 1);
```
fs-eire: we can use `pop()`

lib/backends/wasm/ops/binary-op.ts
```diff
@@ -0,0 +1,73 @@
...
+        fun = '_sub_f32';
+        break;
+      case 'Mul':
+        fun = '_mul_f32';
+        break;
+      case 'Div':
+        fun = '_div_f32';
+        break;
+      case 'Xor':
+        fun = '_xor_f32';
```
fs-eire: _xor_f32 doesn't make sense. this operator should only works with a Boolean type

lib/wasm-binding-core.ts
```diff
@@ -25,8 +25,8 @@ export interface WasmCallArgumentTypeMap {
   int32: number;
   float32: number;
   float64: number;
-  int32ptr: ReadonlyArray<number>|Int32Array|null;
-  float32ptr: ReadonlyArray<number>|Float32Array|null;
+  int32ptr: ReadonlyArray<number>|Uint32Array|Int32Array|null;
+  float32ptr: ReadonlyArray<number>|Int32Array|Uint32Array|Float32Array|null;
```
fs-eire: I think for WASM, all basic types (`int32`, `uint32`, `float`, ...) need to be treated differently (not like javascript there is just a `number`). When used in the context of a specific function, for example, `_add_f32`, an integer type should not work.

src/wasm-build-config.json
```diff
@@ -1,13 +1,23 @@
 {
   "src": [
-    "./wasm-ops/*.cpp"
+    "./wasm-ops/*.cpp",
+    "./wasm-ops/*/*.cpp"
```
fs-eire: does `"./wasm-ops/**/*.cpp"` work here?

src/wasm-build-config.json
```diff
@@ -1,13 +1,23 @@
...
   ],
   "exported_functions": [
     "_malloc",
     "_free",
+	  "_add_f32",
```
fs-eire: remove TABs

src/wasm-ops/binary-op.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "common.h"
+#include "utils/broadcast_utils.h"
+#include "utils/shape_utils.h"
+
+// Core op
+float add_core(const float &a, const float &b) { return a + b; }
+float sub_core(const float &a, const float &b) { return a - b; }
+float mul_core(const float &a, const float &b) { return a * b; }
+float div_core(const float &a, const float &b) { return a / b; }
+float xor_core(const float &a, const float &b) {
```
fs-eire: use `bool` here

src/wasm-ops/binary-op.cpp
```diff
@@ -0,0 +1,100 @@
...
+                    const int32_t &output_rank,
+                    const std::vector<int32_t> &output_dims,
+                    float (*core_op)(const float &, const float &)) {
+  const std::vector<int32_t> &strides1 = ShapeUtils::compute_strides(dims_1);
+  const std::vector<int32_t> &strides2 = ShapeUtils::compute_strides(dims_2);
+  const std::vector<int32_t> &output_strides =
+      ShapeUtils::compute_strides(output_dims);
+
+  for (size_t i = 0; i < output_length; ++i) {
+    auto broadcasted_indices = ShapeUtils::offset_to_indices(output_strides, i);
```
fs-eire: is it possible to avoid creating `std::vector` instances inside the loop?
the size of those output indices should be determined. can we create vectors outside the loop so that those util functions use (instead of create and return) them

src/wasm-ops/utils/broadcast_utils.cpp
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#include "broadcast_utils.h"
+
+std::vector<int32_t> BroadcastUtils::broadcasted_to_original_indices(
+    const std::vector<int32_t> &broadcasted_indices,
+    const std::vector<int32_t> &dims) {
+  const auto rank = static_cast<size_t>(dims.size());
```
fs-eire: I don't see the usage of those `static_cast` necessary... is it the compiler complaining about that?

src/wasm-ops/utils/broadcast_utils.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#pragma once
+
+#include <vector>
+
+class BroadcastUtils {
```
fs-eire: use namespace instead of non-state static class

lib/backends/wasm/ops/binary-op.ts
```diff
@@ -0,0 +1,73 @@
...
+    const outputType = this.resultType ? this.resultType : inputs[0].type;
+    type fourByteTypes = Float32Array|Int32Array|Uint32Array;
+
+    if (inputType === 'float32' || inputType === 'int32' || inputType === 'uint32') {
+      WasmBinding.getInstance().ccall(
+          fun, [inputs[0].numberData as fourByteTypes, 'float32ptr'], [inputs[0].dims.length, 'int32'],
+          [inputs[0].dims, 'int32ptr'], [inputs[1].numberData as fourByteTypes, 'float32ptr'],
+          [inputs[1].dims.length, 'int32'], [inputs[1].dims, 'int32ptr'], [resultData, 'float32ptr', 'out'],
+          [resultData.length, 'int32'], [outputShape.length, 'int32'], [outputShape, 'int32ptr']);
+    } else {
```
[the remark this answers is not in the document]
hariharans29: We will support just f32 first. This is okay as we do that for other operators that consume other numeric types for the WASM backend.

lib/backends/wasm/ops/matmul.ts
hariharans29: Actually it is hard to put this in /lib/ops/matmul but it might make sense to put this into utils

lib/wasm-binding-core.ts
```diff
@@ -233,6 +245,21 @@ export class WasmBinding {
         case 'float32':
           heapF32[offset32] = paramData as number;
           break;
+        case 'boolptr':
+          const boolArray = (paramData as WasmCallArgumentTypeMap['boolptr']);
+          // ReadonlyArray<boolean>
+          if (Array.isArray(paramData)) {
+            const data = heapU8.subarray(offset8, offset8 + boolArray.length);
```
fs-eire: use `set()` should work

lib/backends/wasm/ops/binary-op.ts
```diff
@@ -17,57 +17,75 @@ export class WasmBinaryOp extends BinaryOp {
     if (!outputShape) {
       throw new Error('not broadcastable');
     }
-    const resultDataLength = ShapeUtil.size(outputShape);
-    const resultData = new Float32Array(resultDataLength);
     let fun = '';
+    let binaryOpType = '';
```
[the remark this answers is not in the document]
hariharans29: Put a // TODO:

PR #55, opened 2018-12-27
Add tanh ops for webgl backend, fix sigmoid overflow

NTT123:
Support `tanh` ops for webgl backend. I could write a faster version of `tanh/sigmoid` with the risk of float overflow in `exp(x)` when `x>0` is large.

Instead in this PR, I used  `exp(x - abs(x))` as a replacement for `exp(x)`


fs-eire:
Happy new year! Greetings and thanks for your contributions on ONNX.js @NTT123 

This change is a trade off that scarifies execution speed to cover the edge case. To figure out whether should take this change, we need to answer these questions:
- when will the float overflow happen
   - if it happens, what is the consequences (wrong result/NaN/INF, or exception/crash)
- how frequently this will happen in real usage ( by running real models )

IF the overflow issue seldom happens, or it is not causing big trouble, I would suggest to keep it as-is
IF the overflow issue causes trouble and worth a deal, we can introduce a flag to switch the behavior corresponding.

NTT123:
Just run a few tests (`x=-100 to 100, y = tanh/sigmoid(x)`), it turns out `sigmoid` function has no problem with overflow. 

Meanwhile, the naive `tanh` implementation 

`tanh(x) = { t = exp(2x); return (t-1)/(t+1); }` 

returns wrong result when `x >= 40`. (It returns `0` instead of the correct `1`)

I'll close this PR for now.

PR #72, opened 2019-01-09
Add Unsqueeze ops for webgl backend, refactor cpu backend

NTT123:
I moved part of `CPUUnsqueeze` code to `ShapeUtil` for sharing it with webgl backend.
`WebGLUnsqueeze` uses `WebGLReshape` for doing the real work.


NTT123:
use reshape function as suggested.

PR #73, opened 2019-01-09
Add Squeeze ops for cpu and webgl backends

NTT123:
`Squeeze` ops has similar implementation as `UnSqueeze` ops.


NTT123:
Following @fs-eire's review to use reshape function instead of the object

PR #76, opened 2019-01-14
util: Remove redundant method 

hariharans29:
parseAxis() is the same as getActualAxisFromNegativeValue(). 

Remove the dupe and associated fixes.


fs-eire:
the replacement is not fully completed. see the CI build failure for details

hariharans29:
I missed committing a file or two. It should be good now, I think.

PR #77, opened 2019-01-15
test: optimize data source of test cases

fs-eire:
All test cases were moved out of the repo because of the size concerns. It turns out to be not convenience for feature development since a change usually updates the unittest.

There are 3 types of test cases, the **node**( single node ONNX model, usually used for operator testing ), **op**( .json format ONNX.js unit test) and **model**( the real ONNX model, usually huge in size ). This change updated the data source for these 3 types of test cases:
- node: consume from [**ONNX**](https://github.com/onnx/onnx) repo under folder `deps/onnx/onnx/backend/test/data/node`
- op: consume from this repo under folder `test/data/ops`
- model: consume from [**onnxjs-demo**](https://github.com/Microsoft/onnxjs-demo) under folder `deps/data/data/test/onnx/v7`


tools/test-runner-cli.ts
```diff
@@ -19,10 +19,9 @@ const args = parseTestRunnerCliArgs(process.argv.slice(2));
 logger.verbose('TestRunnerCli.Init.Config', inspect(args));
 
 const TEST_ROOT = path.join(__dirname, '..', 'test');
-const TEST_DATA_ROOT = path.join(__dirname, '..', 'deps', 'data', 'data', 'test');
-const TEST_DATA_NODE_ROOT = path.join(TEST_DATA_ROOT, 'node');
-const TEST_DATA_ONNX_ROOT = path.join(TEST_DATA_ROOT, 'onnx', 'v7');
-const TEST_DATA_OP_ROOT = path.join(TEST_DATA_ROOT, 'ops');
+const TEST_DATA_NODE_ROOT = path.join(__dirname, '..', 'deps/onnx/onnx/backend/test/data/node');
+const TEST_DATA_ONNX_ROOT = path.join(__dirname, '..', 'deps/data/data/test/onnx/v7');
```
hariharans29: I think this can be TEST_DATA_MODEL_ROOT or  TEST_DATA_ONNX_MODEL_ROOT. The naming is little confusing.

PR #78, opened 2019-01-16
util: Remove redundant unnecessary code in BroadcastUtil class

hariharans29:


lib/util.ts
```diff
@@ -84,24 +84,37 @@ export class BroadcastUtil {
...
+      broadcastedIndices: ReadonlyArray<number>, originalShape: ReadonlyArray<number>, originalIndices: number[]) {
+    // NOTE 1: we assume the parameter broadcastedIndices is valid. ie. it should have the same
     // length as the broadcasted shape, and for each dimension the index should
     // not be out of range.
-    const dimOffset = indices.length - originalShape.length;
-    const indicesOriginal = indices.slice(dimOffset);
-    const dimLen = isMatMul ? indicesOriginal.length - 2 : indicesOriginal.length;
-    for (let i = 0; i < dimLen; i++) {
-      indicesOriginal[i] = indices[dimOffset + i] % originalShape[i];
+    // NOTE 2: we assume the parameter originalIndices has the same
```
fs-eire: comments can be put in oneline

lib/util.ts
```diff
@@ -112,31 +125,63 @@ export class BroadcastUtil {
...
-        // traversal indices
-        let rest = i;
-        for (let j = shape.length - 1; j >= 0; j--) {
-          indices[j] = rest % shape[j];
-          rest = Math.floor(rest / shape[j]);
+          outputShape as number[]);
+
+      // both inputs are scalars
+      if (outputShape.length === 0) {
+        c.set(...[], op(a.get(...[]), b.get(...[])));
```
fs-eire: `c.set( op( a.get(), b.get() ) );` should work

lib/util.ts
```diff
@@ -112,31 +125,63 @@ export class BroadcastUtil {
...
+        let isBScalar = false;
+        if (a.shape.length === 0) {
+          valA = a.get(...[]);
+          isAScalar = true;
+        }
+        if (b.shape.length === 0) {
+          valB = b.get(...[]);
+          isBScalar = true;
         }
+        let rest: number;
```
fs-eire: why do we need this `rest` instead of using `i`

lib/util.ts
```diff
@@ -112,31 +125,63 @@ export class BroadcastUtil {
...
-        const indicesB = BroadcastUtil.index(indices, b.shape);
+          if (!isAScalar) {
+            // map outputIndices (which is actually broadcasted) to the originalIndices
+            BroadcastUtil.fillIndex(outputIndices, a.shape, originalIndicesA);
+            // get value
+            valA = a.get(...originalIndicesA);
+          }
+          if (!isBScalar) {
+            BroadcastUtil.fillIndex(outputIndices, b.shape, originalIndicesB);
+            // get value
```
fs-eire: I think we can remove the comment here since `get value` does not helps much

lib/util.ts
hariharans29: We need `rest` as that value gets modified in the inner loop. This is no different from existing implementation, only that rest has been declared just once outside the outer loop...

lib/util.ts
hariharans29: For NOTE 2, I made it single lined. For NOTE 1, it cannot be made a one-liner and I would prefer to keep both separate as they have differing themes.

PR #82, opened 2019-01-24
operator: InstanceNormalization operator for wasm and cpu backends

hariharans29:
This PR is to support InstanceNormalization operator for wasm and cpu backends


lib/backends/cpu/ops/instance-normalization.ts
```diff
@@ -0,0 +1,65 @@
...
+    channelSize *= inputDimensions[i];
+  }
+  const output = new Tensor(x.dims, x.type);
+
+  const X = x.floatData;
+  const Y = output.floatData;
+  const scaleData = scale.numberData;
+  const bData = b.numberData;
+
+  let temp: number;
```
fs-eire: it is better to define those variables inside the for loop.

- no performance concern like objects (`std::vector<>` for example) in C++
- it's more clear that value of those variables is not carried through iterations.

lib/backends/wasm/ops/instance-normalization.ts
```diff
@@ -0,0 +1,51 @@
...
+    WasmBinding.getInstance().ccall(
+        '_instance_normalization_f32', [x.floatData, 'float32ptr'], [y.floatData, 'float32ptr', 'out'],
+        [x.dims[0], 'int32'], [x.dims[1], 'int32'], [channelSize, 'int32'], [scale.floatData, 'float32ptr'],
+        [b.floatData, 'float32ptr'], [this.epsilon, 'float32']);
+
+    return [y];
+  }
+
+  // overriding the checkInputTypes() in the base class because Wasm backend has special type limitations
+  checkInputTypes(inputs: Tensor[]): boolean {
```
fs-eire: can we apply the special type check only? I think typescript keyword `super` can help to call parent class method.

hariharans29: Right now `checkInputTypes()` in the base class also does some other validation other than type validation and hence we need to copy over that snippet of code as well in this method as this method overrides completely the base method. We can make a change later to make it in such a way that only type check is overridden and other kinds of input validation is still common across base and derived classes (hence prevent code duplication). But there are one or two more ops that require this kind of change. So I prefer to do this separately in another change.

lib/backends/cpu/ops/instance-normalization.ts
hariharans29: I see what you mean, and I will keep this in mind for future. I vote we keep this as is for now unless you have any concerns.

hariharans29:
Let's hold off on this change for a little bit - I see that onnxruntime implemented the IN op using Eigen and hence I would like to be consistent with that as much as possible. Even though I see no scope for optimization and I feel the current implementation is good enough, I would like to understand if we directly use onnxruntime code. I will try this tomorrow if possible.

lib/backends/cpu/ops/instance-normalization.ts
fs-eire: we can keep this as is.

hariharans29:
Merging this change - if need be we will re-visit later and use Eigen based implementation if perf is an issue for a real model.

Issue #87, opened 2019-02-07
TypeError: unrecognized operator 'Flatten'


astu9880:
I was able to bypass issue #84 by removing the functino x.size(). However, onnxjs doesn't support another operator I have use of i.e. Flatten. Is there a workaround for the same?

astu9880:
@NTT123 @hariharans29

NTT123:
In pytorch, we can use `view` function as an alternative for `flatten` function . However, 
here is the source code of torch.onnx exporting function for view ops:
https://github.com/pytorch/pytorch/blob/master/torch/onnx/symbolic.py#L440
so, `x = x.view( *sizes) `, can be exported to either `Flatten ops` or `Reshape ops`.

Here we have three options:

  1. modify the above symbolic `view` function to export `reshape` ops only.
  2. use `x = x.view(*size)` cleverly so it falls into the case of exporting `reshape` ops.
  3. implement a `flatten` ops for ONNX.js

fs-eire:
My understanding is, operator `Flatten` only changes the dims of the input tensor like `Reshape`, `Squeeze` and `Unsqueeze` do. If it is this case, it should be easy to implement this operator.

astu9880:
@NTT123 I am unsure about how onnx C functions are working. Can you explain how the 1st and 2nd methods of your suggestion can be executed?

alxmamaev:
Flatten also is using in [ResNet50 version 1](https://github.com/onnx/models/tree/master/models/image_classification/resnet) in ONNX Model Zoo,  so this model is not working in onnx.js

NTT123:
As @fs-eire said, `Flatten` only changes the shape of input tensor. It should be very similar to my previous PR for `Squeeze` here https://github.com/Microsoft/onnxjs/pull/73/files . The only different is to use a new function `squeezeShape` instead of `flattenShape`.

@astu9880 1st and 2nd methods are basically implementing `flatten` function using reshape ops.

fs-eire:
This should have been resolved in #93.

PR #93, opened 2019-02-17
add Flatten ops for webgl and cpu backends

NTT123:
Address missing Flatten issue: https://github.com/Microsoft/onnxjs/issues/87


lib/ops/flatten.ts
```diff
@@ -0,0 +1,38 @@
...
+  initialize(attributes: Attribute): void {
+    this.axis = attributes.getInt('axis', 1);  // default axis is 1
+  }
+
+  checkInputs(inputs: Tensor[]): boolean {
+    if (!inputs || inputs.length !== 1) {
+      return false;
+    }
+
+    if (this.axis < 0 || this.axis >= inputs[0].dims.length) {
```
hariharans29: Can `axis` be legitimately negative ? I have seen some ops that accept negative axis which would mean counting from back from the innermost axis. There is a helper in ShapeUtil to handle negative axis..

lib/util.ts
```diff
@@ -667,6 +667,19 @@ export class ShapeUtil {
     return size;
   }
 
+  /**
+   * Determines the shape of output tensor y = flatten(x, axis)
+   * @param dims - shape of input tensor
+   * @param axis - flatten axis
+   */
+  static flattenShape(dims: ReadonlyArray<number>, axis: number): ReadonlyArray<number> {
+    const total = dims.reduce((x, y) => x * y, 1);
```
hariharans29: Do we really need to reduce dims to know the total size ? We actually already have it from the property `inputs[0].length`. I suggest adding an optional parameter `size` in this function that can be total. If unspecified, it can be computed from `dims` using current logic.

lib/ops/flatten.ts
```diff
@@ -0,0 +1,38 @@
...
+      return false;
+    }
+
```
hariharans29: Probably should return `false` for a scalar tensor (i.e.) if the input tensor has dims.length == 0.

lib/ops/flatten.ts
NTT123: I'm following the Flatten spec. at https://github.com/onnx/onnx/blob/master/docs/Operators.md#Flatten

lib/util.ts
NTT123: I prefer keeping the current function's signature as it is consistent with other shapeutils functions in the `Util` package. Computing `total` can be a bit slower, however the logic is easier to understand.

lib/ops/flatten.ts
NTT123: Added a checking statement for scalar tensor.

Issue #96, opened 2019-02-23
TypeError: unrecognized operator 'Elu'


tms1337:
When trying to load up my model I get

```
TypeError: unrecognized operator 'Elu'
```

I exported my model via pytorch. Does this mean I can't use Elu with onnxjs?

I use the tutorial code for loading

```
const { Tensor, InferenceSession } = require('onnxjs')

const main = async () => {
  const session = new InferenceSession()

  const url = './model-1.onnx'
  await session.loadModel(url)
}

main()
```

Best,

hariharans29:
Resolved by #101

PR #101, opened 2019-03-05
Add ELU activation function

NTT123:
Resolve the issue https://github.com/Microsoft/onnxjs/issues/96 .


lib/backends/webgl/ops/elu.ts
```diff
@@ -0,0 +1,40 @@
...
+  run(inferenceHandler: WebGLInferenceHandler, inputs: Tensor[]): Tensor[] {
+    return WebGLOperatorHelper.run(this, inferenceHandler, inputs);
+  }
+  createProgramInfo(handler: WebGLInferenceHandler, inputs: Tensor[]): ProgramInfo {
+    const outputShape = inputs[0].dims.slice();
+    const shaderSource = `
+      uniform sampler2D A;
+      void main() {
+        float v = texture2D(A, TexCoords).r;
+        gl_FragColor = vec4(v >= 0.0 ? v: (exp(v) - 1.0) * ${this.alpha.toExponential()}); /* float number format */
```
hariharans29: Just curious - Is there a specific reason why you cook the exponential notation of the number in the shader ? (We did not do this for leaky relu, which more or less has very similar logic, so want to know if there is value in adding this to that shader as well)

NTT123: I figured out that the default alpha for for elu is 1.0 which is formatted as integer "1" in the shader source code. This triggers a type error at the shader compiler. 

The default alpha for leaky relu is 0.1, so we didn't have this problem before. 
owever, a custom alpha of integer values such as  1.0, 2.0 could trigger a type error also.

hariharans29: Thanks. Yes we did use float(value) in a lot of places to overcome the type error during shader compilation, it looks like we missed it for leaky relu and we will fix it. 

Butdo you know if there is any difference between using the float constructor (like float (value)) vs your approach ?

NTT123: I can't think of any difference.

PR #105, opened 2019-03-12
WebGL: Fix potential compilation failure in LeakyRelu op

hariharans29:


fs-eire:
could you please rebase the change?

I guess the reason why GitHub cannot show diff correctly is your change is not based on latest master.

hariharans29:
I think it's just the line ending issue again. It is based on latest master. If you see, the change itself has been explicitly highlighted (float())

hariharans29:
Will propose another PR. I had to ditch this fork for various reasons. Since the change is small, I will create a new PR.

PR #106, opened 2019-03-12
WebGL: Fix potential compilation failure in LeakyRelu op

hariharans29:


lib/backends/webgl/ops/leaky-relu.ts
```diff
@@ -19,7 +19,7 @@ export class WebGLLeakyRelu extends LeakyRelu implements WebGLOperator {
       uniform sampler2D A;
       void main() {
         float v = texture2D(A, TexCoords).r;
-        gl_FragColor = vec4(v < 0.0 ? v * ${this.alpha} : v);
+        gl_FragColor = vec4(v < 0.0 ? v * float(${this.alpha}) : v);
```
fs-eire: is `float()` works in compile time or runtime in glsl? if it is the latter then `toExponential()` is a better option

hariharans29: I think we can tackle this solution across all the shaders (many shaders need the answer to that investigation)

PR #107, opened 2019-03-12
operator: Support Clip op for cpu, wasm, and webgl backends

### LINKED ISSUE #92: Clip operation is not supported
If you trying to export and load MobilenetV2 (like in this [implementation](https://github.com/tonylins/pytorch-mobilenet-v2/blob/master/MobileNetV2.py))
You will get error: `unrecognized operator 'Clip'`
This error is except because in MobilenetV2 are using [ReLU6](https://pytorch.org/docs/stable/nn.html#torch.nn.ReLU6) instead ReLU.

hariharans29:
Resolve #92


lib/backends/wasm/ops/clip.ts
```diff
@@ -0,0 +1,34 @@
...
+import {WasmBinding} from '../../../wasm-binding';
+import {WasmInferenceHandler} from '../inference-handler';
+
+export class WasmClip extends Clip {
+  run(inferenceHandler: WasmInferenceHandler, inputs: Tensor[]): Tensor[] {
+    const result = new Tensor(inputs[0].dims, inputs[0].type);
+    const size = result.floatData.length;
+    if (inputs[0].type === 'float32') {
+      WasmBinding.getInstance().ccall(
+          '_clip_f32', [inputs[0].floatData, 'float32ptr'], [result.floatData, 'float32ptr', 'out'], [size, 'int32'],
```
hariharans29: No use of templatizing as `unary-op` in WASM backend as there is not much to templatize (no core implementation here). In fact, the overhead in dealing with types and attributes parsing (which is much more prevalent than in `binary-op` case makes it an unattractive option. So, keep the kernels separate even for `unary-op` like implementations

src/wasm-ops/clip.h
```diff
@@ -0,0 +1,21 @@
...
+#include "common.h"
+
+extern "C" {
+void clip_f32(void *);
+// Expand for other supported data types for `clip`
+}
+
+// Core implementation of the op
+template <typename T>
+void clip_imp(const T *input, T *output, const int32_t length, const float min,
```
hariharans29: No use of templatizing like `binary_imp` as in binary ops there is a lot to templatize on (for example - the bradcasting logic, etc.). Here - there is only a single for loop to templatize on and it becomes cumbersome to deal with attributes in `unary-op` case. So, only templatizing on different `types` of the op and instantiating as necessary.

lib/backends/webgl/ops/clip.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  createProgramInfo(handler: WebGLInferenceHandler, inputs: Tensor[]): ProgramInfo {
+    const outputShape = inputs[0].dims.slice();
+    const shaderSource = `
+      const float min = float(${this.min});
+      const float max = float(${this.max});
+      uniform sampler2D A;
+      void main() {
+        float v = texture2D(A, TexCoords).r;
+        gl_FragColor = vec4(v < min ? min : (v > max) ? max : v);
```
NTT123: There is `clamp` function in WebGL specs at https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/clamp.xhtml . This function has the same semantic as Clip function here.

hariharans29: Thanks a lot. Will take a look.

src/wasm-ops/clip.h
```diff
@@ -0,0 +1,21 @@
...
+              const float max) {
+  for (size_t i = 0; i < length; ++i) {
+    const auto &val = input[i];
+    output[i] = (val < min) ? min : (val > max) ? val : val;
```
fs-eire: output[i] = (val < min) ? min : (val > max) ? **`max`** : val;

src/wasm-ops/clip.h
fs-eire: I think it is OK for now. If we are going to have 10+ unary-op implementations it will be the time to consider this problem.

src/wasm-ops/clip.h
hariharans29: Thanks for catching this. Will fix.

hariharans29:
> Generally looks good to me. there is one small bug in WASM implementation.
> 
> About WASM:
> Compare WASM vs. javascript on element-wise unary op for performance:
> 
> * possible perf gain: avoid float32<->float64 convert for each element (because javascript `number` is float64, and the underlying data is usually `Float32Array`)
> * possible drawback: extra cost in interop (`ccallSerialize()` and `ccallDeserialize()`)
> 
> And this is also a JIT(V8)-vs-emcc battle, so the perf result may be very different in different browser version.
> 
> Due to our previous perf data (as we tried `Relu` before) in resnet50 there is no performance gain from implementation of simple element-wise unary op in WASM. But it is possible that WASM implementation can be faster than javascript. So in my opinion it's a good idea to have WASM unary-op implementation.

Another thing to consider for JS vs WASM unary-ops is the actual operation being performed on each element. For example, let's consider `exp(x)` - I do not know how slow exp(x) is for JS when compared to C++. My hypothesis is that (I do not have proof yet) for certain large number of elements, the interop cost + `float32 <-> float64` + c++ element-wise op performance <= pure JS element-wise op.

src/wasm-ops/clip.h
hariharans29: This is weird, I am not sure, how this is not bieng caught by node tests. I tried addign unit tests, but I realized that our test data sits in some toher repo making it hard to add tests :(

src/wasm-ops/clip.h
hariharans29: I think we can potentially support all unary-ops in WASM. Even then I would find it hard to templatize the core implementation as each has separate attributes. Since there is no native "attribute" object in our WASM layer, it becomes really hard to templatize this.

src/wasm-ops/clip.h
fs-eire: op test is in the repo and node tests are not. do you want to add test case into node tests?

fs-eire:
> Generally looks good to me. there is one small bug in WASM implementation.
> About WASM:
> Compare WASM vs. javascript on element-wise unary op for performance:
> 
> possible perf gain: avoid float32<->float64 convert for each element (because javascript number is float64, and the underlying data is usually Float32Array)
> possible drawback: extra cost in interop (ccallSerialize() and ccallDeserialize())
> 
> And this is also a JIT(V8)-vs-emcc battle, so the perf result may be very different in different browser version.
> Due to our previous perf data (as we tried Relu before) in resnet50 there is no performance gain from implementation of simple element-wise unary op in WASM. But it is possible that WASM implementation can be faster than javascript. So in my opinion it's a good idea to have WASM unary-op implementation.
> 
> Another thing to consider for JS vs WASM unary-ops is the actual operation being performed on each element. For example, let's consider exp(x) - I do not know how slow exp(x) is for JS when compared to C++. My hypothesis is that (I do not have proof yet) for certain large number of elements, the interop cost + float32 <-> float64 + c++ element-wise op performance <= pure JS element-wise op.

It is actually `interop cost` + `WASM c++ element-wise op performance` <= `float32 <-> float64` + `pure JS element-wise op`.

For float32 tensor, the data is stored in `Float32Array`. For javascript, accessing an element in that typed array will cause a `float32 -> float64`, and write an element back to the typed array will cause a `float64 -> float32`. 

The real underlying math function is always native so that if they are same function it should not make difference. however `exp(double)`(happened in js) and `exp(float32)`(happened in wasm) are expected to have difference.

src/wasm-ops/clip.h
hariharans29: I had not resolved this op in the `wasm` backend and so it was passing (despite the bug) all tests by falling back to the  `cpu` backend. I added `wasm` native clip in the operators resolve function and this caught the bug (as expected)

PR #108, opened 2019-03-14
lib: update operator resolve

fs-eire:
This change updates the behavior how to resolve a `Graph.Node` to an `Operator` in a session's initialization stage.

- Change the IR version check to support IR version 4
- Enable opset version check
- Replace the switch-case based resolve to a data driven resolve
- Allows opset version selector (currently support exact match and range match, eg. `7`, `7+`)


lib/backend.ts
```diff
@@ -32,10 +33,9 @@ export interface SessionHandler {
   /**
    * Resolves the operator from the name; backend specific
```
hariharans29: nit: I think the comment would be - resolve the operator from the name and the opset version ?

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -0,0 +1,100 @@
...
+  ['GlobalAveragePool', '', '1+', () => new WebGLGlobalAveragePool()],
+  ['GlobalMaxPool', '', '1+', () => new WebGLGlobalMaxPool()],
+  ['Greater', '', '7+', () => new binaryOps.WebGLBinaryOp(NUMBER_TYPES, binaryOps.glslGreater(), undefined, 'bool')],
+  ['Identity', '', '1+', () => new unaryOps.WebGLUnaryOp(NUMBER_TYPES, unaryOps.glslIdentity())],
+  ['ImageScaler', '', '1+', () => new WebGLImageScaler()],
+  ['LeakyRelu', '', '6+', () => new WebGLLeakyRelu()],
+  ['Less', '', '7+', () => new binaryOps.WebGLBinaryOp(NUMBER_TYPES, binaryOps.glslLess(), undefined, 'bool')],
+  ['Log', '', '6+', () => new unaryOps.WebGLUnaryOp(FLOAT_TYPES, unaryOps.glslLog())],
+  ['MatMul', '', '1+', () => new WebGLMatMul()],
+  ['MaxPool', '', '1-7', () => new WebGLMaxPool()],  // TODO: support new attributes for MaxPool-8 and MaxPool-10
```
hariharans29: Developer-centric question - I am guessing the way we support different versions of an op (where there is majority overlap across versions) is we just extend the current working class and override appropriate methods so as to align with the new spec ?

lib/opset.ts
```diff
@@ -0,0 +1,67 @@
...
+  version: number;
+}
+
+export declare namespace OpSet {
+  interface OperatorConstructor {
+    (node: Graph.Node): Operator;
+  }
+
+  /**
+   * Domain of an opset, it can be an empty string(default value, represent for ai.onnx), or 'ai.onnx.ml'
```
hariharans29: can the domain be ai.onnx OR default value (empty string) to represent the ai.onnx domain ?

lib/opset.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (node.opType === opType) {  // operator type matches
+      for (const opset of opsets) {
+        // opset '' and 'ai.onnx' are considered the same.
+        if (opset.domain === domain || (opset.domain === 'ai.onnx' && domain === '')) {  // opset domain found
+          if (matchSelector(opset.version, versionSelector)) {
+            return opConstructor(node);
+          }
+        }
+      }
+      throw new TypeError(`cannot resolve operator '${opType}'`);
```
hariharans29: can error thrown also mention requested opset version of the operator cannot be resolved ?

lib/backends/cpu/op-resolve-rules.ts
```diff
@@ -0,0 +1,90 @@
...
+import {CpuUnsqueeze} from './ops/unsqueeze';
+
+export const CPU_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
+  ['Abs', '', '6+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.abs)],
+  ['Acos', '', '7+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.acos)],
+  ['Add', '', '7+', () => new CpuBinaryOp(NUMBER_TYPES, (e1, e2) => (e1 + e2))],
+  ['And', '', '7+', () => new CpuBinaryOp(['bool'], (e1, e2) => (e1 && e2))],
+  ['ArgMax', '', '1+', () => new CpuArgMax()],
+  ['Asin', '', '7+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.asin)],
+  ['Atan', '', '7+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.atan)],
```
hariharans29: How were the ResolveRules created for the very first time? How can we make sure it is accurate? Do we need to maintain any script(s) used to generate them ?

lib/backend.ts
fs-eire: will fix

lib/backends/webgl/op-resolve-rules.ts
fs-eire: there are 2 ways.
one is to extend the existing implementation, _if the new version is compatible with the old one according to the spec_. (eg. add a new optional attribute, whose default value is the same of the old behavior)
one is to have different implementation, _if the new version has a different behavior, or it is necessary to do so due to performance concern_.

lib/opset.ts
fs-eire: empty string means 'ai.onnx' as in domain

lib/backends/cpu/op-resolve-rules.ts
fs-eire: this is a pain point. one possible way is to grab more test cases from other place.

meanwhile, I am writing a script to generate the operator support table from data (to replace /docs/operators.md). that helps but still not a guarantee of 100% accuracy.

hariharans29:
At a high level, the changes look okay. 

Looking at the build failures it looks to me that there might be issues concerning the array of `ResolveRule` that have been generated for each backend (or some bug in matchSelector() and/or resolveOperator() which I am sure can be ironed out). My only concern is the general accuracy of these set of `ResolveRule` for each backend.

fs-eire:
> At a high level, the changes look okay.
> Looking at the build failures it looks to me that there might be issues concerning the array of ResolveRule that have been generated for each backend (or some bug in matchSelector() and/or resolveOperator() which I am sure can be ironed out). My only concern is the general accuracy of these set of ResolveRule for each backend.

I think the final solution is to make sure we have a good test coverage. Maybe we can leverage some test cases from onnxruntime.

lib/backends/cpu/op-resolve-rules.ts
```diff
@@ -0,0 +1,90 @@
...
+  ['AveragePool', '', '7+', () => new CpuAveragePool()],
+  ['BatchNormalization', '', '7+', () => new CpuBatchNormalization()],
+  ['Ceil', '', '6+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.ceil)],
```
hariharans29: Also, how can we ensure potentially a bad commit that could have conflicting operator resolves possible ?

For example, Ceil 6+ could be mapped to an op, let's say Ceil 10 has has a different op implementation, but the developer forgets to change `6+` to `6-9`. In that case, multiple op resolves are possible.

One way is to run a script at check-in time but we will have to maintain that...

lib/backends/cpu/op-resolve-rules.ts
```diff
@@ -0,0 +1,91 @@
...
+  ['Relu', '', '6+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.relu)],
+  ['Reshape', '', '5+', () => new CpuReshape()],
+  ['Sigmoid', '', '6+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.sigmoid)],
+  ['Sin', '', '7+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.sin)],
+  ['Slice', '', '1+', () => new CpuSlice()],
+  ['Softmax', '', '1+', () => new CpuSoftmax()],
+  ['Sqrt', '', '6+', () => new unaryOps.CpuUnaryOp(NUMBER_TYPES, unaryOps.sqrt)],
+  ['Squeeze', '', '1+', () => new CpuSqueeze()],
+  ['Sub', '', '7+', () => new CpuBinaryOp(NUMBER_TYPES, (e1, e2) => (e1 - e2))],
+  ['Sum', '', '6+', () => new CpuSum()],  // TODO: support multidirectional broadcast for Sum-8
```
hariharans29: Do we need to throw exception when it looks like broadcasting support maybe required in the implementation (since it support 6+) ?

fs-eire: the exception is thrown anyway. currently it will be thrown due to checkInput failure. the only difference is the error message

hariharans29:
There are a couple of nits left out from being resolved/addressed.
1) comment update in backend.ts 
2) update error message in opset.ts to indicate that the opset version of the operator cannot be resolved (as opposed to the generic - operator could not be resolved which is might be misleading to anyone who looks at the code base and finds implementation for that operator (from an earlier opset))

PR #121, opened 2019-04-03
test: support multi version node cases

fs-eire:
This change update script tools/build.ts to copy v7-v10 node test cases and allow different node test version for test runner.

BrowserStack builds requires #120 as dependency to pass


tools/build.ts
```diff
@@ -80,6 +81,51 @@ logger.info('Build.SubModules', '(2/2) Fetching submodules... DONE');
...
+  if (cleanInstall) {
+    fs.emptyDirSync(TEST_DATA_NODE);
+  }
+  logger.info('Build.TestData', `(1/3) Cleaning folder... ${cleanInstall ? 'DONE' : 'SKIPPED'}`);
+  // Step 2: copy node tests for different version
+  logger.info('Build.TestData', '(2/3) Copy tests...');
+  [['v7', '5af210ca8a1c73aa6bae8754c9346ec54d0a756e'],   // rel-1.2.3
+   ['v8', '5cc146270945f0d007b140fb59a892a60ba69f49'],   // rel-1.3.0
+   ['v9', '4e67414849122d3df78bd72cee5717f90e715d12'],   // rel-1.4.1
+   ['v10', 'b22041c3f16cf8bcca9ed93982d6ffdf6ebf3746'],  // master
```
hariharans29: Is the general idea to freeze a commit id after every onnx release ? For example after onnx locks down v10, we freeze the commit id for that and we keep rolling forward the commit id of master (for v11, and so on) ?

tools/test-runner-cli.ts
```diff
@@ -35,11 +37,14 @@ const json = stripJsonComments(jsonWithComments, {whitespace: true});
 const whitelist = JSON.parse(json) as Test.WhiteList;
 logger.verbose('TestRunnerCli.Init', `Loading whitelist... DONE`);
 
-const DEFAULT_BACKENDS: ReadonlyArray<TestRunnerCliArgs.Backend> = ['cpu', 'wasm', 'webgl'];
+// The default backends and opset version lists. Those will be used in suite tests.
+const DEFAULT_BACKENDS: ReadonlyArray<TestRunnerCliArgs.Backend> =
+    args.env === 'node' ? ['cpu', 'wasm'] : ['cpu', 'wasm', 'webgl'];
+const DEFAULT_OPSET_VERSIONS: ReadonlyArray<number> = [10, 9, 8, 7];
```
hariharans29: Do we need to keep upgrading this after each onnx release too ? If so, I think we better document this...

tools/build.ts
fs-eire: yes that is the idea.

tools/test-runner-cli.ts
fs-eire: the answer is 'yes' ideally. but considering the fact that node test cases are rarely modified ( usually the change is to add cases to that folder), we can do this as demand.

there is another option: only support the stable ONNX version. This means we should only support v10 after ONNX v1.5 release (which will happen very soon though). thus we only update when a ONNX release happen

PR #133, opened 2019-04-12
test: determine file preload based on file size

fs-eire:
Update the test runner's behavior on file cache.
- Previous:
  use file cache only one node test cases (used in suite tests)
- Current:
  use file cache when file size <= 4MB

This helps the test speed especially when in `--perf` mode.


fs-eire:
Replaced by #136

PR #135, opened 2019-04-15
test: make file cache async

fs-eire:
This change allows the TestRunner's file cache to be access asynchronously.

This helps to reduce the loading time when running suite test, which add 1000+ entries into the file cache.


fs-eire:
Replaced by #136

PR #139, opened 2019-04-23
webgl: texture leak fix

fs-eire:
This change fixes the texture leak bug that happened in some operators.


lib/backends/webgl/ops/conv.ts
```diff
@@ -169,18 +166,18 @@ export class WebGLConv extends Conv {
...
+    const kLayout = inferenceHandler.createTextureLayoutFromShape(
         adjustedKernelShape, 4, [adjustedKernelShape[0], adjustedKernelShape[1] * 4], {breakAxis: 1});
 
     let bLayout: TextureLayout|undefined;
     const rank = outputShape.length;
 
     const inputLayouts = [im2colLayout, kLayout];
     if (inputs.length === 3) {
-      bLayout = inferenceHandler.createBasicTextureLayout(inputs[2].dims.slice());
+      bLayout = inferenceHandler.createTextureLayoutFromShape(inputs[2].dims.slice());
```
hariharans29: There seem to be a lot of unnecessary slicing (copying) of `dims` in general - as long as the methodd accepts const dims and guarantees that any given dims will not be mutated - can we avoid creating copies unnecessarily ? Granted -it should be only adding to the warm up overhead, but it still kind of stands out...

fs-eire: thanks for figuring out this. actually there are quite some unnecessary calls to `Array.prototype.slice` in a lot of places in the code base.

I think it worth another separated change to optimize that

lib/backends/webgl/inference-handler.ts
```diff
@@ -5,105 +5,151 @@ import {InferenceHandler} from '../../backend';
...
+   *   Creates a texture data object associated with the given tensor.
+   * @param tensor the tensor with data to upload
+   */
+  getOrCreateTextureData(tensor: Tensor, layout?: TextureLayout) {
+    let td = this.getTextureData(tensor.dataId);
     if (!td) {
       Logger.verbose('InferenceHandler', `Creating new TextureData for dims: [${tensor.dims}]`);
       if (!layout) {
-        layout = this.createBasicTextureLayout(tensor.dims.slice());
+        layout = this.createTextureLayoutFromShape(tensor.dims.slice());
```
hariharans29: Do we need this `slice()` ?

lib/backends/webgl/inference-handler.ts
```diff
@@ -5,105 +5,151 @@ import {InferenceHandler} from '../../backend';
...
-      Logger.verbose('InferenceHandler', `Retrieving Tensor from cache for:[${td.unpackedShape}]`);
+      this.textureDataCache.set(tensorId, td);
     }
-    return tensor;
   }
+
+  /**
+   * Create a TextureLayout object from a tensor. If a related texture data is found, returns the cached texture layout.
+   */
   getOrCreateTextureLayout(tensor: Tensor, channels = 1, unpackedShape?: ReadonlyArray<number>): TextureLayout {
```
hariharans29: Nit: This name is slightly misleading. I would expect that `getOrCreateTexturelayout()` to only focus on Layout  creation. There is a step that looks into the cache preceeding it. Any possible better names for this function ?

lib/backends/webgl/inference-handler.ts
```diff
@@ -5,105 +5,151 @@ import {InferenceHandler} from '../../backend';
...
-    const td = this.getTextureData(tensor);
+    const td = this.getTextureData(tensor.dataId);
     if (td) {
       return td;
     }
-    return this.createBasicTextureLayout(
+    return this.createTextureLayoutFromShape(
         channels === 1 ? tensor.dims.slice() : getPackedShape(tensor.dims.slice()), channels, unpackedShape);
```
hariharans29: Do we need to make all the copies using `slice()` ?

lib/backends/webgl/ops/conv.ts
hariharans29: Oh yeah, I seem to have this comment already - too many `slice` calls :)

lib/backends/webgl/inference-handler.ts
```diff
@@ -5,105 +5,151 @@ import {InferenceHandler} from '../../backend';
...
+      layout: TextureLayout, dataType: Tensor.DataType, data?: Tensor.NumberType, tensor?: Tensor,
+      usage?: Encoder.Usage): TextureData {
+    Logger.verbose('InferenceHandler', `Creating TextureData: layout:[${JSON.stringify(layout)}]`);
+    const texture = this.textureHelper.createTextureFromLayout(dataType, layout, data, usage);
+    return this.createTextureDataFromTexture(layout, dataType, texture, tensor);
+  }
+
+  /**
+   * Create a TextureData object, using the given texture.
+   * This function does not create new texture. Usually used in scenarios using texture sharing. (eg. Reshape)
```
hariharans29: Can you expand on `texture sharing` for my understanding please ?

hariharans29:
Overall looks good to me - I focussed on the "core" WebGL related changes. I skimmed through the related WebGL operator changes that needs to align with the base change itself. I just need to understand the concept of `createSharedTextureData()` and its uses. I ll chat with you offline and we can merge this.

lib/backends/webgl/inference-handler.ts
fs-eire: close those duplicated comments related to `slice()`

lib/backends/webgl/inference-handler.ts
fs-eire: This is what the function is doing: try to get (find from cache) the texture layout of the given tensor; if failed to find then create a new one.. so from the naming perspective I cannot find a better name for it. This is the same for `getOrCreateTextureData`.

lib/backends/webgl/inference-handler.ts
fs-eire: A `WebGLTexture` object represents a texture in GPU.

A `TextureData` object represents a corresponding texture to a tensor. It contains a `WebGLTexture` object, a reference to the tensor and the texture layout.

A `Tensor.Id` object represents a unique data unit, regardless where the data is hold (it can be in memory, ie. TypedArray, or in GPU, ie. a texture).

A `Tensor` object represent a value node in the model graph. It's one of those:
  - a model input
  - a model initializer
  - a model output
  - an intermedia value through the model execution
A `Tensor` object can have multiple way to connect to the data that its tensorId mapped to:
  - a cached result (TypedArray instance)
  - a data provider function ( a function that returns a TypedArray instance )
  - an async data provider function (this is not supported yet)

Logically, `WebGLTexture` objects should be 1:1 mapping to `Tensor.Id` objects, and `TextureData` objects should be a 1:1 mapping to `Tensor` objects.

Some operators, for example, `Reshape`, will not modify any data of its input. In this case, we are creating a new `Tensor` object with different `dims`, using the same `Tensor.Id`. This is the sharing scenario. We extend this scenario with the equivalent concepts in webgl.

PR #141, opened 2019-04-23
update version to 0.1.6

fs-eire:


package-lock.json
```diff
@@ -1,9 +1,29 @@
 {
   "name": "onnxjs",
-  "version": "0.1.5",
+  "version": "0.1.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
```
hariharans29: Seems like unusual amount of changes in dependencies. Is this expected ?

fs-eire: package-lock.json is generated by `npm install`. Those changes in package.json are expected.

PR #148, opened 2019-05-11
graph: bugfix for initializer

### LINKED ISSUE #147: Error: load converted models keras2onnx
Error: invalid name for initializer: W when load model
this first conv layer in model

fs-eire:
- allows implicit inputs in initializers
- allows inputs with multiple None dimensions

Fixes #147

### DIFF
--- a/lib/session.ts
+++ b/lib/session.ts
@@ -204,11 +204,9 @@ export class Session {
     }
 
     for (let i = 0; i < expectedDims.length; ++i) {
-      if (expectedDims[i] !== actualDims[i]) {
-        if (!noneDimSupported || expectedDims[i] !== 0) {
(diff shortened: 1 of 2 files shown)


lib/session.ts
```diff
@@ -203,23 +203,10 @@ export class Session {
...
-          if (noneDimSeen) {
-            return false;
-          }
-
-          noneDimSeen = true;
-        }
-
-        // Not a 'None' dimension. Data shape mis-match.
-        else {
+        if (!noneDimSupported || expectedDims[i] !== 0) {
```
hariharans29: maybe just merge the 2 ifs  expectedDims[i] !== actualDims[i] && (... || ...) ?

Issue #116, opened 2019-03-29
Android `webgl` backend outputs are incorrect and inconsistent between devices


gnsmrky:
On Google Pixel 3, `Pad 'Reflection'` with `webgl` backend outputs zeros.
On Huawei Mate 20 Pro, `Pad 'Reflection'` is correct, but `Conv2D` output is not.

Is there a requirement for `webgl` on Android devices?  

I've created a quick test page at [ONNX Op test page (https://gnsmrky.github.io/pytorch-fast-neural-style-onnxjs/test_ops.html)](https://gnsmrky.github.io/pytorch-fast-neural-style-onnxjs/test_ops.html).  Can someone have a quick look at the above issues?  I am trying to find a combination that works across different Android devices using `webgl` backend.

fs-eire:
@gnsmrky I tried to reproduce this issue using Android emulator on PC. However the issue is not shown in the emulator. I think I need a real device to test and debug (I will get one in couple of days).

gnsmrky:
@fs-eire yes I think webgl is very hardware dependent. The webgl in Android emulator should be using either cpu-emulated webgl, or the Intel webgl backend.  So was trying to find out if there is a minimum comformant requirement when running mobile.

Would suggest Pixel devices as target device as it is being developed by Google.  It is just odd that Conv2d does not work on Huawei device.  Gonna try to find more Android devices for testing Conv2d.

Issue #145, opened 2019-05-07
webgl extension


ShuangLiu1992:
with the latest release I'm getting this error with chrome on android. (samsung note 9, samsung s6)

[.WebGL-0xcc479c00]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: GL_BLEND with floating-point color attachments requires the EXT_float_blend extension

ShuangLiu1992:
is there anyway to force webgl1 context since this extension doesn't seem to be supported on android?

ShuangLiu1992:
I can switch the context to webgl1 with 
` onnx.backend.webgl.contextId = 'webgl';`

then it gives me 
```
GlContextFactory 2019-05-07T14:33:10.547Z|failed to create WebGLContext using contextId 'webgl' or 'experimental-webgl'. Error: Error: both float32 and float16 TextureType are not supported
:8000/onnx.js:1 [30;43mw WebGLBackend 2019-05-07T14:33:10.549Z|Unable to initialize WebGLBackend. Error: WebGL is not supported
```

fs-eire:
It looks like GL_BLEND not working on Samsung devices (it works on Google Pixel, LG, Sony, OnePlus and so on. I don't have any Huawei, Xiaomi so not sure about those)

What makes the situation worse is, the WebGL extension availability is not detectable. The [spec](https://www.khronos.org/registry/webgl/extensions/EXT_float_blend/) indicates that `EXT_color_buffer_float` should enable BLEND in "supported systems". Samsung devices support this extension but it is apparently not among the "supported systems". This means there is no easy way to detect whether GL_BLEND is supported.

For WebGL1, it is literally no float support on those Samsung devices. In WebGL1, `OES_texture_float` is required to create float32 format texture, and if it is not available, fallback to `OES_texture_half_float` to use float16. If neither is available there is no way to use webgl backend in ONNX.js. On Samsung devices those 2 extensions are not available in webgl1.

ShuangLiu1992:
yea, I thought onnxjs was using webgl1 on ios but seems like it is just using wasm backend.
samsung phones are quite popular in eruope and tensorflowjs does support them (also supports synchronous inference).

onnxjs is a great framework and has lots of promises but at its current state we will have to go back to converting onnx models to keras then tensorflow js :c

fs-eire:
[...]

Issue #127, opened 2019-04-08
yolo demo on mobile


ShuangLiu1992:
yolo demo on mobile doesn't work
device: samsung note 8 with latest chrome

fs-eire:
Thanks for the feedback. I think this may caused by the same reason as #116. I'm working on a change in texture manager which hopefully help.

ShuangLiu1992:
fixed in #149

PR #149, opened 2019-05-11
webgl: fix issues on Android devices

fs-eire:
This PR contains changes to a lot of files. Major changes are list as below:
 - refactor webgl backend
 - support GLSL ES 3.0
 - enable check for GL_BLEND
 - update test cases, test whitelist, test runner and debug output

This change should fix issue #116, #127, #145


lib/backends/backend-webgl.ts
```diff
@@ -21,13 +21,13 @@ export class WebGLBackend implements Backend, WebGLOptions {
...
-  forceUint8Reads = false;
+  matmulMaxBatchSize?: number;
 
   initialize(): boolean {
     try {
       this.glContext = createWebGLContext(this.contextId);
-      if (!this.glContext.floatDownloadEnabled) {
-        this.forceUint8Reads = true;
+      if (typeof this.matmulMaxBatchSize !== 'number') {
+        this.matmulMaxBatchSize = 16;
```
hariharans29: is this because this was the default value previously ? I mean there is no data to choose this value right ?

lib/backends/webgl/glsl-function-inliner.ts
```diff
@@ -1,59 +1,54 @@
...
+                         const tokens = s.trim().split(' ');
+                         if (tokens && tokens.length === 2) {
+                           return {type: tokens[0], name: tokens[1]};
+                         }
+                         return null;
+                       })
+                       .filter(v => v !== null);
+    inlineDefs[match[2]] = {params, body: match[4]};
+  }
+  for (const name in inlineDefs) {
```
hariharans29: nice refactor!

lib/backends/backend-webgl.ts
fs-eire: this is because it's the previous default value.

PR #153, opened 2019-05-20
remove ndarray from dependencies

fs-eire:
`ndarray` and its related libraries were introduced in the early stage of ONNX.js project. They are easy to use and powerful. However, there are out-of-dated and generate npm warnings.

Only a few amount of ndarray functions were in use: `set()`, `get()`, `assign()` and `matrixProduct()`. Those are replaced by CPU backend implementations.

Size of onnx.min.js reduced 26kB (452kB -> 426kB)


lib/api/tensor-impl.ts
```diff
@@ -67,18 +66,17 @@ export class Tensor implements TensorInterface {
       if (dim >= this.dims[idx]) {
         throw new RangeError(`Input index array dims don't match the tensor dims.`);
       }
-      flatIndices += idx < indexArray.length - 1 ? dim * this.dims.slice(idx + 1).reduce((a, b) => a * b) : dim;
```
hariharans29: So previously - we were doing the reduce every time instead of pre-computing strides at init time ? 

Currently, I am guessing when we `get()` from internalTensor, we will fetch the value with lesser overhead (i.e.) using strides ?

lib/api/tensor-impl.ts
```diff
@@ -97,17 +95,12 @@ export class Tensor implements TensorInterface {
...
     });
 
     if (typeof value === 'boolean') {
-      this.data[flatIndices] = value ? 1 : 0;
-    } else if (typeof value === 'string') {
-      this.data[flatIndices] = value;
-    } else if (ArrayBuffer.isView(this.data)) {
-      this.data.set([value], flatIndices);
```
hariharans29: btw - do we need to couple API change with ndarray removal ?

lib/backends/cpu/ops/matmul.ts
```diff
@@ -16,71 +14,153 @@ export class CpuMatMul extends MatMul {
...
+  return y;
+}
+
+/**
+ * perform matrix multiply on C = alpha * A * B + beta * C
+ * @param A data of tensor A, whose shape is [M,K] or [K,M] (if transA)
+ * @param B data of tensor B, whose shape is [K,N] or [N,K] (if transB)
+ * @param C data of tensor C, whose shape is [M,N]
+ */
+export function matMul2d(
```
hariharans29: Are these helpers coming from Lotus ? Should we check-in some explicit unit tests for these (can come in a separate PR). Currently, I guess they are being implicitly tested through model usage.

lib/util.ts
```diff
@@ -164,40 +160,43 @@ export class BroadcastUtil {
    * @param a The input tensor A
    * @param b The input tensor B
    * @param op The operator lambda function
+   * @param implace Whether to write the result back to A.
    * @returns The result tensor, or undefined if input not broadcastable.
    */
-  static calc(a: ndarray, b: ndarray, op: (a: number, b: number) => number): ndarray|undefined {
-    const outputShape = BroadcastUtil.calcShape(a.shape, b.shape);
+  static calc(
+      a: Tensor, b: Tensor, op: (a: string|number, b: string|number) => (string | number), implace: boolean,
```
hariharans29: nit: you mean `inplace` ?

lib/backends/cpu/ops/argMax.ts
```diff
@@ -10,13 +10,13 @@ export class CpuArgMax extends ArgMax {
...
-export function argMax(x: Tensor, axis: number, keepdims: number): Tensor {
+export function argMax(x: Tensor, axis: number, keepdims: boolean): Tensor {
   const rank = x.dims ? x.dims.length : 1;
   axis = ShapeUtil.parseAxis(axis, rank);
-  const outputDims = ReduceUtil.calcReduceShape(x.dims.slice(0), [axis], 1);
+  const outputDims = ReduceUtil.calcReduceShape(x.dims, [axis], true);
   const X = x.data;
   const Y = new Int32Array(ShapeUtil.size(outputDims));
-  const blockSize = axis >= x.dims.length ? 1 : ShapeUtil.size(x.dims.slice(axis + 1));
+  const blockSize = axis + 1 >= x.dims.length ? 1 : ShapeUtil.sizeFromDimension(x.dims, axis + 1);
```
hariharans29: dumb question 1: if `axis +1 > x.dims.length`, shouldn't it be an error case ?

dumb question 2: can't we directly call sizeFromDimension(x.dims, axis + 1) and it should return 1 if axis +1 == x.dims.length  ?

lib/api/tensor-impl.ts
fs-eire: yes, you are right. current internal tensor uses cache for strides (it will be only calculated once)

but this is still slow. in general, get() and set() are much less efficient than using the typed array directly.

lib/api/tensor-impl.ts
fs-eire: ndarray removal does not affect any API.

lib/backends/cpu/ops/argMax.ts
fs-eire: good catch! I checked the code and found it can be simplified to `sizeFromDimension(x.dims, axis + 1) `

PR #155, opened 2019-05-25
utils: fix shape inference for Conv

fs-eire:
This change fixes a bug that may cause a wrong output shape for `Conv` operator, when dilation is not 1.


lib/util.ts
```diff
@@ -1117,47 +1121,49 @@ export class PoolConvUtil {
...
+    const dkernel = dilation * (kernel - 1) + 1;
     if (autoPad && autoPad !== 'NOTSET') {
       switch (autoPad) {
         case 'VALID':
           pads[padHeadIndex] = 0;
           pads[padTailIndex] = 0;
-          return Math.floor(((inSize - kernel) / stride) + 1);
+          return Math.floor(((inSize - dkernel) / stride) + 1);
         case 'SAME_LOWER':
           const legacyTargetSize1 = (inSize + stride - 1) / stride;
```
hariharans29: enforce dilation == 1  for SAME_UPPER and SAME_LOWER to keep parity with Lotus

lib/util.ts
```diff
@@ -1117,47 +1121,49 @@ export class PoolConvUtil {
...
           const padNeeded1 = (legacyTargetSize1 - 1) * stride + kernel - inSize;
           pads[padHeadIndex] = Math.floor((padNeeded1 + 1) / 2);
           pads[padTailIndex] = padNeeded1 - pads[padHeadIndex];
-          return Math.floor(((inSize + padNeeded1 - kernel) / stride) + 1);
+          return Math.floor(((inSize + padNeeded1 - dkernel) / stride) + 1);
```
hariharans29: can be kernel itself for now since dkernel makes it confusing slightly.

lib/util.ts
```diff
@@ -1117,47 +1121,49 @@ export class PoolConvUtil {
...
         case 'SAME_UPPER':
           const legacyTargetSize = (inSize + stride - 1) / stride;
           const padNeeded = (legacyTargetSize - 1) * stride + kernel - inSize;
           pads[padHeadIndex] = Math.floor(padNeeded / 2);
           pads[padTailIndex] = padNeeded - pads[padHeadIndex];
-          return Math.floor(((inSize + padNeeded - kernel) / stride) + 1);
+          return Math.floor(((inSize + padNeeded - dkernel) / stride) + 1);
```
hariharans29: same as above

lib/util.ts
```diff
@@ -1084,7 +1086,9 @@ export class PoolConvUtil {
     // Add batch size and number of channels of output
     const outputDims = [inputDims[0], inputDims[1]];
 
-    PoolConvUtil.computeShapeHelper(isGlobalOperator, inputDims, outputDims, strides, kernelShape, pads, autoPad);
+    PoolConvUtil.computeShapeHelper(
+        isGlobalOperator, inputDims, outputDims, strides, new Array<number>(kernelShape.length).fill(1), kernelShape,
```
hariharans29: comment here TODO...

PR #158, opened 2019-05-31
webgl: refactor uint8_encode by graph transformer

fs-eire:
this change moves the uint8_encode from a standalone class to a custom operator. this helps to keep consistency on the use of WebGLContext.


fs-eire:
cancel this PR as it breaks several tests. should fix those first

PR #198, opened 2020-07-21
webgl: support grouped convolution

fs-eire:


lib/backends/webgl/ops/conv.ts
```diff
@@ -231,18 +231,15 @@ export class WebGLConv extends Conv {
     if (group === 1 && (channels === 1 || (shape[2] * shape[3]) % channels === 0)) {
       return kernel;
     }
-    const strides = ShapeUtil.computeStrides(shape);
```
hariharans29, line 234: Do we need a check to ensure: `input_channels/group == filter_channels` somewhere ?

fs-eire: yes. this check is done at https://github.com/fs-eire/onnxjs/blob/grouped-conv/lib/ops/conv.ts#L35-L40

PR #206, opened 2020-08-13
build(deps): update packages to latest

fs-eire:
- package `@types/webgl2` is removed because it conflicts with typescript 3.7+, which added type definitions for WebGL2.
- package `worker-loader` v3.0 introduced API changes in options, renaming 'name' to 'publicPath'.
- package `karma-safari-launcher` is replaced by `karma-safari-applescript-launcher` due to the old launcher does not work on Safari v12+
- latest Safari does not support `Uint8Array.from( number[] )`; use `new Uint8Array()` instead.


lib/api/tensor-impl.ts
```diff
@@ -26,7 +26,7 @@ export class Tensor implements TensorInterface {
         this.data = Float32Array.from(data as number[]);
       } else if (type === 'bool') {
         // convert boolean[] to Uint8Array
```
hanbitmyths, line 28: Add a comment to use new instead of from method here. When safari supports Uint8Array.from, this can be changed for consistency with other lines.

PR #207, opened 2020-08-13
src: operator Upsample

fs-eire:
This change adds support for operator Upsample in opset[7-8] backend cpu/webgl.


lib/backends/cpu/ops/upsample.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export class CpuUpsample extends Upsample {
+  run(inferenceHandler: CpuInferenceHandler, inputs: Tensor[]): Tensor[] {
+    const xDims = inputs[0].dims;
+    const yDims = xDims.map((dim, i) => Math.floor(dim * this.scales[i]));
+    const y = new Tensor(yDims, inputs[0].type);
+    if (this.mode === 'nearest') {
+      upsampleNearest(inputs[0].data, y.data, xDims, yDims, this.scales);
+    } else {
+      upsampleLinear(inputs[0].data, y.data, xDims, yDims, this.scales, this.roi);
```
hanbitmyths, line 16: Seems like roi is not used. What is this for?

lib/backends/cpu/ops/upsample.ts
```diff
@@ -0,0 +1,156 @@
...
+  const dx2 = new Array<number>(outputWidth);
+
+  for (let y = 0; y < outputHeight; ++y) {
+    let inY = getOriginalCoordinate(y, heightScale);
+    yOriginal.push(inY);
+    inY = Math.max(0, Math.min(inY, inputHeight - 1));
+
+    const inY1 = Math.min(Math.floor(inY), inputHeight - 1);
+    const inY2 = Math.min(inY1 + 1, inputHeight - 1);
+    dy1[y] = Math.abs(inY - inY1);
```
hanbitmyths, line 102: Put these two lines into if-else below?

lib/backends/cpu/ops/upsample.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  for (let x = 0; x < outputWidth; ++x) {
+    let inX = getOriginalCoordinate(x, widthScale);
+    xOriginal.push(inX);
+    inX = Math.max(0, Math.min(inX, inputWidth - 1));
+
+    inX1[x] = Math.min(Math.floor(inX), inputWidth - 1);
+    inX2[x] = Math.min(inX1[x] + 1, inputWidth - 1);
+
+    dx1[x] = Math.abs(inX - inX1[x]);
```
hanbitmyths, line 122: Same here. Put these two lines into if-else below?

lib/backends/webgl/ops/upsample.ts
```diff
@@ -0,0 +1,193 @@
...
+    const glsl = getGlsl(handler.session.backend.glContext.version);
+
+    const outputPitches = new Array<number>(dim);
+    const inputPitches = new Array<number>(dim);
+    let precalculatedPitches = `
+      int output_pitches[${dim}];
+      int input_pitches[${dim}];
+      `;
+    for (let d = dim - 1; d >= 0; d--) {
+      outputPitches[d] = (d === dim - 1) ? 1 : outputPitches[d + 1] * outputShape[d + 1];
```
hanbitmyths, line 29: Set outputPitches[dim - 1] and inputPitches[dim - 1] before this loop to avoid unnecessary comparison.

test/data/ops/upsample.jsonc
```diff
@@ -0,0 +1,53 @@
...
+              4
+            ],
+            "dims": [1, 1, 4, 6],
+            "type": "float32"
+          }
+        ]
+      }
+    ]
+  }
+]
```
hanbitmyths, line 53: No test for linear case?

lib/backends/cpu/ops/upsample.ts
fs-eire: Explanation of why -
 - There are different opset (7, 9, 10) of `Upsample` in ONNX spec, in this change we only implement [7-8]. Our goal is to support all opset version but we have priorities so this time the change is for [7-8] only.
- in opset v10, `Upsample` is deprecated. It's replaced by `Resize`, which is designed to support both upsample and downsample. It also introduced more attributes, in which `roi` is one of them.
- definition of Upsample does not have a roi input, however, Resize does. the cpu backend implementation is basically referencing implementation of onnxruntime, which uses a general implementation that works for Upsample's each opset version, and Resize as well.

I think I can remove the roi here for now and add it back when implementing other opsets.

lib/backends/webgl/ops/upsample.ts
fs-eire: I did this change locally but then I realize that this requires to duplicate the 4 lines of code in the for body, which makes the code hard to read. considering this condition expression does very little impact to performance, and this code only happens in initialization, it does not have enough benefit to balance the lost of readability; so I will not change this place.

test/data/ops/upsample.jsonc
fs-eire: added multiple test cases.

PR #208, opened 2020-08-14
src: fix web-worker

fs-eire:
This change fixes web workers in ONNX.js:
- #206 introduced an initialization failure of web worker because `worker-loader` v3.0 behaves differently. Fixes include:
   - split parameter `name` to `publicPath` and `filename`. (publicPath is omitted as its default value is "./")
   - `exports` is no longer a function. need to use `exports.default`
- the number of workers should be determined by the environment. Specifically, we use [navigator.hardwareConcurrency](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorConcurrentHardware/hardwareConcurrency) to determine the best worker number. If it is not available (Usually in Safari) we set it to 1.


lib/backends/backend-wasm.ts
```diff
@@ -60,3 +53,32 @@ export class WasmBackend implements Backend, WasmOptions {
     }
   }
 }
+
+function defaultNumWorkers(): number {
+  if (typeof navigator !== 'undefined' && navigator) {
+    // by default use ([navigator.hardwareConcurrency / 2] - 1) workers
+    if (typeof navigator.hardwareConcurrency === 'number') {
+      return Math.max(Math.ceil(navigator.hardwareConcurrency / 2) - 1, 0);
```
hanbitmyths, line 61: If navigator.hardwareConcurrency doesn't exist, is the default number of workers 0 or 1? I mean, what's the basis that two behaviors exists here? Is it possible to make the logic simple to return 0 or 1 without checking Safari case here?

fs-eire: Safari browser ( on Apple devices ) does not expose API `navigator.hardwareConcurrency`, but it has the availability to launch web workers. use `1` as number of web worker helps to leverage the multithread capability and it's not too aggressive.

If it is not Safari browser, this should not happen to any modern browser. This is most likely because the browser is too old to support web worker.

PR #227, opened 2020-09-08
src: allow operators to access graph in initialization

fs-eire:
By accessing the graph ( already resolved and finalized ) in operator's initialization, it is possible to access the tensor value of the operator's input, if it's the graph's initializer or const value. This will enable implementation of upcoming operators that want to deal with constant inputs.

### DIFF
--- a/lib/backend.ts
+++ b/lib/backend.ts
@@ -32,16 +32,16 @@ export interface SessionHandler {
 
   /**
    * Resolves the operator from the name and opset version; backend specific
-   * @param node
-   * @param opsets
+   * @param node the node to resolve
+   * @param opsets a list of opsets that exported from the model
+   * @param graph the completely initialized graph
    */
+  resolve(node: Graph.Node, opsets: ReadonlyArray<OpSet>, graph: Graph): Operator;
 
-  resolve(node: Graph.Node, opsets: ReadonlyArray<OpSet>): Operator;
   /**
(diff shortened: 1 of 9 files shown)


lib/backend.ts
```diff
@@ -32,16 +32,16 @@ export interface SessionHandler {
 
   /**
    * Resolves the operator from the name and opset version; backend specific
-   * @param node
-   * @param opsets
+   * @param node the node to resolve
+   * @param opsets a list of opsets that exported from the model
+   * @param graph the completely initialized graph
    */
+  resolve(node: Graph.Node, opsets: ReadonlyArray<OpSet>, graph: Graph): Operator;
```
hanbitmyths, line 39: NIT: not that different, but make graph as optional?

lib/operators.ts
```diff
@@ -3,10 +3,11 @@
 
 import {Attribute} from './attribute';
 import {InferenceHandler} from './backend';
+import {Graph} from './graph';
 import {Tensor} from './tensor';
 
 export interface Operator {
-  initialize(attributes: Attribute): void;
+  initialize(attributes: Attribute, node: Graph.Node, graph: Graph): void;
```
hanbitmyths, line 10: Same here. node and graph as optional?

hanbitmyths (approved):

hanbitmyths (approved):

PR #228, opened 2020-09-08
implement Resize operator

fs-eire:
This change uses unified implementation that works with `Upsample` v7, `Upsample` v9, `Resize` v10 and `Resize` v11, in both CPU and WebGL backends. Depend on #227 to build.


hanbitmyths (approved):
Regarding test cases failure, I think this review has a dependency with interface change, right? Please, double-check and merger after all tests are passed.

lib/ops/upsample.ts
```diff
@@ -3,33 +3,104 @@
...
+
+    this.useNearest2xOptimization = (this.opset < 11) ?
+        true :
+        (this.mode === 'nearest' && this.coordinateTransformMode === 'asymmetric' && this.nearestMode === 'floor');
+
+    if (this.opset > 10) {
+      this.roiInputIdx = 1;
+      this.scalesInputIdx = 2;
+      this.sizesInputIdx = 3;
+    } else if (this.opset === 9) {
```
debkbanerji, line 70: I ran into an issue when trying to use this when this.opset was 10. In my particular use case, adding `|| this.opset === 10` caused the inference to run as expected, but I am not an expert regarding this codebase.

PR #246, opened 2020-12-17
webgl: fix 'Gather' with 0-rank indices

fs-eire:
This change fixes 'Gather' with 0 rank indices.


hanbitmyths (approved):
Looks good to me

Issue #177, opened 2020-01-08
LogSoftmax operator not supported


yissachar:
I've exported an ONNX model from PyTorch using opset version 7 (the lowest version the PyTorch export supports), and it includes a LogSoftmax operator, which results in the following error when I try to use it in ONNX.js:

```
TypeError: cannot resolve operator 'LogSoftmax' with opsets: ai.onnx v7
```
This version of LogSoftmax has been available since version 1 of ONNX so it would be nice to support it.

LeonardoEmili:
Any chance of this being supported?

PR #254, opened 2021-01-03
Add LogSoftmax for CPU

hayashikun:
I need LogSoftmax, and I added LogSoftmax for the CPU backend. (related to #177)

This is the first time for me, so please let me confirm some points.

Regarding the `test-suite-whitelist.jsonc`,

- I added some tests written in [onnx/onnx](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/logsoftmax.py).
Is it enough?

- I added a commnet line in `ops` because there is not `log-softmax.jsonc` in [onnxjs-demo](https://github.com/microsoft/onnxjs-demo/tree/data/data/test/ops).
Should I prepare test cases and send PR to onnxjs-demo?

Best Regards,

### DIFF
--- a/lib/backends/cpu/op-resolve-rules.ts
+++ b/lib/backends/cpu/op-resolve-rules.ts
@@ -34,6 +34,7 @@ import * as unaryOps from './ops/unary-op';
 import {CpuUnaryOp} from './ops/unary-op';
 import {CpuUnsqueeze} from './ops/unsqueeze';
 import {CpuUpsample, CpuUpsampleV9} from './ops/upsample';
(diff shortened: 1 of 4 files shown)


fs-eire:
it looks like the build failed on pre-commit check. could you please run `npm run format` and check it again?

hayashikun:
Thank you for your advice!
CI looks to work well except BrowserStack, which has failed in other PRs.

iiSeymour:
Would be godd to see this make a release (#267).

iiSeymour:
I have a `wasm` and `webgl` implementation I can add also.

PR #269, opened 2021-03-06
return from test if on mac

Tixxx:


test/unittests/backends/webgl/test_pack_unpack.ts
```diff
@@ -154,6 +154,15 @@ describe('#UnitTest# - pack - Tensor pack', () => {
     inferenceHandler = sessionhandler.createInferenceHandler();
   });
   const testDataSet = getTestData();
+
+  // TODO support WebGl 1.0 to run it on Mac
```
fs-eire, line 158: use `createWebGLContext()` from webgl-context-factory.ts to perform a webgl version check

fs-eire (changes requested):

test/unittests/backends/webgl/test_pack_unpack.ts
Tixxx: updated

test/test-runner.ts
```diff
@@ -230,10 +230,42 @@ async function initializeSession(
 /**
  * run a single model test case. the inputs/outputs tensors should already been prepared.
  */
-export async function runModelTestSet(context: ModelTestContext, testCase: Test.ModelTestCase): Promise<void> {
-  Logger.verbose('TestRunner', `Start to run test data from folder: ${testCase.name}`);
+// TODO fix the reshape ops to be compatible with webgl1
```
xzhu1900, line 233: reshape and flatten ops?

xzhu1900 (approved):

test/test-runner.ts
Tixxx: although flatten is implemented with reshape, it would be good to look into flatten as well for corner cases. updated.

xzhu1900 (approved):

fs-eire (approved):

PR #272, opened 2021-03-12
Tix/concat packed

Tixxx:
packed version of concat

### DIFF
--- a/test/unittests/backends/webgl/test_concat_packed.ts
+++ b/test/unittests/backends/webgl/test_concat_packed.ts
@@ -117,6 +117,7 @@ interface TestData {
   inputTextureShape: number[];
   outputTextureShape: number[];
   expectedOutput: Float32Array;
+  // If empty, the test will use auto-generated data.
   rawInput?: Float32Array;
 }
 function getTestData(): TestData[] {
--- a/test/unittests/backends/webgl/test_concat_packed.ts
+++ b/test/unittests/backends/webgl/test_concat_packed.ts
@@ -4,8 +4,9 @@ import {expect} from 'chai';
 
 import {Attribute} from '../../../../lib/attribute';
 import {Backend, InferenceHandler, SessionHandler} from '../../../../lib/backend';
+import {WebGLBackend} from '../../../../lib/backends/backend-webgl';
 import {WebGLInferenceHandler} from '../../../../lib/backends/webgl/inference-handler';
-import {WebGLPackedConcat} from '../../../../lib/backends/webgl/ops/concat_packed';
+import {WebGLConcat} from '../../../../lib/backends/webgl/ops/concat';
 import {Profiler} from '../../../../lib/instrument';
 import {Tensor} from '../../../../lib/tensor';
 import {ShapeUtil} from '../../../../lib/util';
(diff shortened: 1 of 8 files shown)


test/unittests/backends/webgl/test_concat_packed.ts
```diff
@@ -0,0 +1,279 @@
...
+    // },
+    // {
+    //   elementCount: 64,
+    //   axis: 1,
+    //   inputShape: [2, 2, 4, 4],
+    //   outputShape: [2, 2, 4, 4],
+    //   inputTextureShape: [2, 8],
+    //   outputTextureShape: [16, 4],
+    //   useGeneratedOutput: true,
+    // },
```
xzhu1900, line 277: Should these be re-enabled?

xzhu1900:
Shall we also add this packed concat kernel into `onnxjs\lib\backends\webgl\op-resolve-rules.ts`, so model contains concat op can be resolved with this kernel?

test/unittests/backends/webgl/test_concat_packed.ts
Tixxx: was trying to see how the max CI would react to concat, looks like it didnt like that either...  added the check to skip on webgl1 and re-enabled other tests.

Tixxx:
> Shall we also add this packed concat kernel into `onnxjs\lib\backends\webgl\op-resolve-rules.ts`, so model contains concat op can be resolved with this kernel?

I was hesitant to do so since we now no longer have 1-to-1 mapping from onnx op to webgl kernel. I changed the old kernel api to be a wrapper of 2 different implementations, in that way we can dynamically switch to packed or unpacked depending on the inputs, pack mode and etc.. Since it now keeps 2 versions of the op impl, the memory overhead is unknown. will need to check that.

duli2012:
> > Shall we also add this packed concat kernel into `onnxjs\lib\backends\webgl\op-resolve-rules.ts`, so model contains concat op can be resolved with this kernel?
> 
> I was hesitant to do so since we now no longer have 1-to-1 mapping from onnx op to webgl kernel. I changed the old kernel api to be a wrapper of 2 different implementations, in that way we can dynamically switch to packed or unpacked depending on the inputs, pack mode and etc.. Since it now keeps 2 versions of the op impl, the memory overhead is unknown. will need to check that.

The dynamic approach offers more flexibility at the expense of runtime overhead (which should be negligible in most cases though). The key is whether there're scenarios in which runtime info like inputs is required to decide which kernel to pick. Or, on the contrary, static info like kernel attributes is enough. For the latter, the kernel selection can be done at model loading time, e.g. in op-resolve-rules.ts.

xzhu1900 (approved):
:shipit:

Tixxx:
> > > Shall we also add this packed concat kernel into `onnxjs\lib\backends\webgl\op-resolve-rules.ts`, so model contains concat op can be resolved with this kernel?
> > 
> > 
> > I was hesitant to do so since we now no longer have 1-to-1 mapping from onnx op to webgl kernel. I changed the old kernel api to be a wrapper of 2 different implementations, in that way we can dynamically switch to packed or unpacked depending on the inputs, pack mode and etc.. Since it now keeps 2 versions of the op impl, the memory overhead is unknown. will need to check that.
> 
> The dynamic approach offers more flexibility at the expense of runtime overhead (which should be negligible in most cases though). The key is whether there're scenarios in which runtime info like inputs is required to decide which kernel to pick. Or, on the contrary, static info like kernel attributes is enough. For the latter, the kernel selection can be done at model loading time, e.g. in op-resolve-rules.ts.

Yea the memory overhead is very small in this case since the Operator class is very light-weight, and there's no difference in runtime perf. I think it's inevitable to have a differentiation based on input size, i haven't measured it in-depth myself. TFJS doesn't use the packed concat implementation for 1d tensors so I'm using the similar logic here for concat.

PR #274, opened 2021-03-17
Conv pack kernel

duli2012:
Adding a Conv pack kernel


lib/env.ts
```diff
@@ -15,6 +15,7 @@ class EnvironmentImpl implements ENV {
   public readonly platform = platform;
 
   public debug = false;
+  public packMode = true;
```
xzhu1900, line 18: >public packMode = true; [](start = 2, length = 23)

just a heads-up: we removed and renamed this boolean. This may result in some merge conflict.

lib/backends/webgl/ops/conv.ts
```diff
@@ -113,6 +113,8 @@ export class WebGLConv extends Conv {
 
     const rank = outputShape.length;
     const im2colDims = WebGLConv.calcIm2ColDims(xshape, kshape, outputShape, 4);
+    console.log('outputShape= ', outputShape);
+    console.log('im2colDims= ', im2colDims);
```
xzhu1900, line 117: remove?

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -47,6 +49,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['Clip', '', '6-10', () => new WebGLClip()],
   ['Concat', '', '4+', () => new WebGLConcat()],
   ['Conv', '', '1+', () => new WebGLConv()],
+  ['ConvPacked', '', '1+', () => new WebGLConvPacked()],
```
xzhu1900, line 52: >ConvPacked [](start = 4, length = 10)

we need to find a good way to resolve kernels with both pack and unpack version supported. One way is through webGL session handler's resolve method, another is through individual kernels. 

The one through session handler is more systematic, but as it calls into a generic `resolveOperator` function, which is shared between all backends, it makes it harder to make any webGL backend specific changes. One possible solution is to change `resolveOperator` to take a function pointer instead, and have each backends decide how to resolve ops specifically. However, this would require a lot of change. The other option is more light-weighted. Similar to `concat`, it queries session handler and append different source code for fragment shader. 

Personally I feel option one should be the goal for long run as it is more centralized and clean, but may have a lower priority given current planning. I'm also good with option 2 for now too.

xzhu1900: Same for MatmulPack below.

---
In reply to: [596286236](https://github.com/microsoft/onnxjs/pull/274#discussion_r596286236) [](ancestors = 596286236)

lib/backends/webgl/program-manager.ts
```diff
@@ -105,11 +105,12 @@ ${fragShaderScript}
...
   bindOutput(td: TextureData): void {
+    const width = td.width;
+    const height = td.height;
     Logger.verbose(
         'ProrgramManager',
-        `Binding output texture to Framebuffer: w/h=${td.width}/${td.height}, shape=${td.shape}, type=${
-            td.tensor.type}`);
-    this.glContext.attachFramebuffer(td.texture, td.width, td.height);
+        `Binding output texture to Framebuffer: w/h=${width}/${height}, shape=${td.shape}, type=${td.tensor.type}`);
+    this.glContext.attachFramebuffer(td.texture, width, height);
```
xzhu1900, line 113: revert?

lib/backends/webgl/webgl-context.ts
```diff
@@ -103,7 +103,7 @@ export class WebGLContext {
   readTexture(
       texture: WebGLTexture, width: number, height: number, dataSize: number, dataType: Encoder.DataType,
       channels: number): Encoder.DataArrayType {
-    const gl = this.gl;
+    const gl = this.gl as WebGL2RenderingContext;
```
xzhu1900, line 106: >const gl = this.gl as WebGL2RenderingContext [](start = 4, length = 44)

Looks like this branch is based off the old pack_share branch. There are a handful changes made to that branch afterwards before merging. Maybe it's not a bad idea to try rebase.

test/data/ops/matmulpack.jsonc
```diff
@@ -0,0 +1,31 @@
+[
```
xzhu1900, line 1: nit: consider rename it to matmul_pack.jsonc for consistancy

lib/env.ts
duli2012: fixed

lib/backends/webgl/ops/conv.ts
duli2012: done

lib/backends/webgl/op-resolve-rules.ts
duli2012: right. let's have a discussion offline.

lib/backends/webgl/program-manager.ts
duli2012: done.

lib/backends/webgl/webgl-context.ts
duli2012: fixed.

test/data/ops/matmulpack.jsonc
duli2012: done.

lib/backends/webgl/op-resolve-rules.ts
duli2012: Fixed. Now it's aligned with concat and reshape pack version.

lib/util.ts
```diff
@@ -1080,7 +1080,6 @@ export class PoolConvUtil {
     if (!autoPad) {
       return;
     }
-
     if (pads.length !== 2 * (inputDims.length - 2)) {
```
xzhu1900, line 1083: >     [](start = 0, length = 4)

Nit: blank change. Consider revert this file.

lib/backends/webgl/ops/unpack.ts
```diff
@@ -6,7 +6,6 @@ import {getGlsl} from '../glsl-source';
 import {WebGLInferenceHandler} from '../inference-handler';
 import {ProgramInfo, RunData, WebGLOperator} from '../types';
 import {getCoordsDataType} from '../utils';
-
 import {getChannels, unpackFromChannel} from './packing_utils';
```
xzhu1900, line 9: >import {getChan [](start = 0, length = 15)

nit: revert this file

lib/backends/webgl/ops/matmul.ts
```diff
@@ -6,10 +6,20 @@ import {Tensor} from '../../../tensor';
 import {BroadcastUtil} from '../../../util';
 import {WebGLInferenceHandler} from '../inference-handler';
 import {ProgramInfo, RunData, WebGLOperator} from '../types';
+import {WebGLMatMulPacked} from './matmul-pack';
 
 export class WebGLMatMul extends MatMul implements WebGLOperator {
+  packedImpl: WebGLMatMulPacked;
```
xzhu1900, line 12: > packedImpl: WebGLMatMulPacked; [](start = 1, length = 31)

It's a bit weird to have an unpacked WebGLMatMul class contains a Packed MatMul object as member variable. Theoratically, those two implementations should be peer instead of one containing another. Maybe we can add a layer of abstraction, called WebGLMatMul, and rename the current unpacked implementation WebGLMatMulUnpacked (align with WebGLMatMulPacked). In WebGLMatMul, depending on inferenceHandler.session.pack, choose different kernel?

lib/backends/webgl/ops/matmul-pack.ts
```diff
@@ -0,0 +1,64 @@
...
+export class WebGLMatMulPacked extends MatMul implements WebGLOperator {
+  run(inferenceHandler: WebGLInferenceHandler, inputs: Tensor[]): Tensor[] {
+    return inferenceHandler.run(this, inputs);
+  }
+  createProgramInfo(handler: WebGLInferenceHandler, inputs: Tensor[]): ProgramInfo {
+    const hasBias = inputs.length > 2;
+    const processBias = hasBias ? `value += vec4(getBias(a[0]*2).xx, getBias(a[0]*2).yy);` : ``;
+    const aShape = inputs[0].dims;
+    const bShape = inputs[1].dims;
+    const outputShape = BroadcastUtil.calcShape(aShape, bShape, true);
```
xzhu1900, line 19: I think our current implementation doesn't cover cases when one of the input tensor needs broadcast? If so, we may want to add a TODO here for future re-visiting.

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -68,6 +72,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['Less', '', '7+', () => new binaryOps.WebGLBinaryOp(NUMBER_TYPES, binaryOps.glslLess(), undefined, 'bool')],
   ['Log', '', '6+', () => new unaryOps.WebGLUnaryOp(FLOAT_TYPES, unaryOps.glslLog())],
   ['MatMul', '', '1+', () => new WebGLMatMul()],
+  ['MatMulPacked', '', '1+', () => new WebGLMatMulPacked()],
```
xzhu1900, line 75: remove

lib/backends/webgl/glsl-coordinate-lib.ts
```diff
@@ -1216,4 +1224,27 @@ export class CoordsGlslLib extends GlslLib {
...
   }
+
+  /**
+   * Produces a packed value getter function for the name and rank given
+   * If a transpose is set proper offsetToCoords mapping will be used
+   * @param name name of the function
+   * @param rank rank of the input
+   * @param transpose whether or not should generate a transpose variation
+   */
+  protected getPackedValueFrom(varName: string, rank: number, width: number, height: number, transpose: boolean):
```
xzhu1900, line 1235: >getPackedValueFrom [](start = 12, length = 18)

Do we need this function? Or can we use existing function in this library.

lib/backends/webgl/ops/matmul-pack.ts
duli2012: done

lib/util.ts
duli2012: done.

lib/backends/webgl/ops/unpack.ts
duli2012: done.

lib/backends/webgl/ops/matmul.ts
duli2012: fixed.

lib/backends/webgl/glsl-coordinate-lib.ts
duli2012: Do you mean getA()? the diff beteen getA() and A_Pack() is A_Pack() takes an array as inputs.

xzhu1900: Do you need an array as function input? For example, for pack, we pass in 1-4 parameters to getA() depending on the input ranks. I think that's also the case here?

---
In reply to: [613794313](https://github.com/microsoft/onnxjs/pull/274#discussion_r613794313) [](ancestors = 613794313)

lib/backends/webgl/ops/matmul-pack.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    const rank = outputShape.length;
+    const arank = aShape.length;
+    const brank = bShape.length;
+    const sharedDim = aShape[aShape.length - 1];
+    const shaderSource = `
+      vec4 process(int indices[${rank}]) {
+          int a[${arank}];
+          int b[${brank}];
+          bcastMatmulIndices_A(indices, a);
```
xzhu1900, line 32: >bcastMatmulIndices_A [](start = 10, length = 20)

will this work for broadcaset matmul?

lib/backends/webgl/ops/matmul-pack.ts
```diff
@@ -0,0 +1,64 @@
...
+    const processBias = hasBias ? `value += vec4(getBias(a[0]*2).xx, getBias(a[0]*2).yy);` : ``;
+    const aShape = inputs[0].dims;
+    const bShape = inputs[1].dims;
+    const outputShape = BroadcastUtil.calcShape(aShape, bShape, true);
+
+    if (!outputShape) {
+      throw new Error('Can\'t use matmul on the given tensors');
+    }
+    const rank = outputShape.length;
+    const arank = aShape.length;
```
xzhu1900, line 25: >arank [](start = 10, length = 5)

nit: consider using aRank and bRank

lib/backends/webgl/webgl_utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import {WebGLInferenceHandler} from './inference-handler';
+import {WebGLReshapePacked} from './ops/reshape-packed';
+
+export function reshape(inferenceHandler: WebGLInferenceHandler, input: Tensor, shape: number[]): Tensor {
+  const op = new WebGLReshapePacked();
+  const newShape = new Tensor([shape.length], 'int32');
+  for (let i = 0; i < shape.length; i++) {
+    newShape.data[i] = shape[i];
+  }
+  return op.run(inferenceHandler, [input, newShape])[0];
```
xzhu1900, line 11: Do we still need this change?

lib/backends/webgl/glsl-coordinate-lib.ts
duli2012: Not entirely impossible to use getA(). it will need a bit clunky code to deal following code:
 for (int k=0; k<((${sharedDim}+1)/2); ++k) {
              a[${arank - 1}] = k;
              b[${brank - 2}] = k;
              value += _A_Pack(a).rrbb * _B_Pack(b).rgrg;
              value += _A_Pack(a).ggaa * _B_Pack(b).baba;
  }
  
It's not quite the same as Pack. For pack, there are four cases for last two dims, but here it's more complicated. Make sense?

lib/backends/webgl/ops/matmul-pack.ts
duli2012: Not yet.

lib/backends/webgl/webgl_utils.ts
duli2012: no. good catch.

lib/backends/webgl/ops/matmul-pack.ts
duli2012: Will fix it in another PR.

xzhu1900:
Do you plan to keep this PR or have another new PR together with the texture cache change?

lib/backends/webgl/ops/conv-pack.ts
```diff
@@ -0,0 +1,117 @@
...
+import {WebGLConv} from './conv';
+import {WebGLIm2ColPacked} from './im2col-pack';
+import {WebGLMatMulPacked} from './matmul-pack';
+import {WebGLReshapePacked} from './reshape-packed';
+
+export class WebGLConvPacked extends Conv {
+  unpackedImpl: WebGLConv;
+  constructor() {
+    super();
+    this.unpackedImpl = new WebGLConv();
```
xzhu1900, line 20: nit: it is a bit weird to have a packed conv including a unpacked conv impl as member variable. Maybe we can add a comment here and also rename it to fallbackImpl?

lib/backends/webgl/ops/conv-pack.ts
```diff
@@ -0,0 +1,117 @@
...
+        'Conv',
+        `autpPad:${this.autoPad}, dilations:${this.dilations}, group:${this.group}, kernelShape:${
+            this.kernelShape}, pads:${this.pads}, strides:${this.strides}`);
+
+    const outputShape = WebGLConv.calcOutputShape(xshape, kshape, this.dilations, this.pads, this.strides);
+    const im2col = new WebGLIm2ColPacked(outputShape, kshape, this.dilations, this.pads, this.strides);
+    const matmul = new WebGLMatMulPacked();
+    const reshape = new WebGLReshapePacked();
+    // shape for kernel reshape
+    const shape = new Tensor([2], 'int32');
```
xzhu1900, line 60: this can be replaced as 

`const shape= new Tensor([2], 'int32', undefined, undefined, new Float32Array([kshape[0],kshape[1] * kshape[2] * kshape[3]]));`

lib/backends/webgl/ops/conv-pack.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    // reshape kernel
+    const runDataKernelReshape = reshape.createRunData(inferenceHandler, this.programInfo[1], [inputs[1], shape]);
+    inferenceHandler.checkAndUpdateTextureForm(this.artifacts[1], runDataKernelReshape);
+    programManager.run(this.artifacts[1], runDataKernelReshape);
+    const kernelReshaped = runDataKernelReshape.outputTextureData.tensor;
+
+    // run matmul
+    const hasBias = (inputs.length === 3);
+    if (this.artifacts.length === 2) {
```
xzhu1900, line 88: Maybe we should add an assertion instead of an if to ensure at this point, the previous artifacts are created?

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -47,6 +48,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['Clip', '', '6-10', () => new WebGLClip()],
   ['Concat', '', '4+', () => new WebGLConcat()],
   ['Conv', '', '1+', () => new WebGLConv()],
+  ['ConvPacked', '', '1+', () => new WebGLConvPacked()],
```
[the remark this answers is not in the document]
xzhu1900: Now I see the op tests for packed conv,  I think it's ok to keep this op here. But I think eventually, we want to keep this list lean and clean and test both modes in our test. We can run all our op tests twice: one with pack mode and one with unpack mode, and both should have correct result.

lib/backends/webgl/ops/conv-pack.ts
```diff
@@ -0,0 +1,117 @@
...
+    const matmulOutput = runDataMatmul.outputTextureData.tensor;
+
+    // reshape output
+    const outputShapeTensor = new Tensor([outputShape.length], 'int32');
+
+    for (let i = 0; i < outputShape.length; i++) {
+      outputShapeTensor.data[i] = outputShape[i];
+    }
+
+    if (this.artifacts.length === 3) {
```
xzhu1900, line 107: same as above, at this point, this.artifacts.length needs to be 3. We should add an assert here.

xzhu1900 (approved):

lib/backends/webgl/op-resolve-rules.ts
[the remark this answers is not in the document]
duli2012: no, we should remove this line and merge op tests for both packed  and unpacked conv.

lib/backends/webgl/ops/conv-pack.ts
duli2012: done

lib/backends/webgl/ops/conv-pack.ts
duli2012: done.

lib/backends/webgl/ops/conv-pack.ts
duli2012: done

lib/backends/webgl/op-resolve-rules.ts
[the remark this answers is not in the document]
duli2012: refactored the code a little bit.

lib/backends/webgl/ops/conv-pack.ts
duli2012: part of the refactoring aforenoted.

lib/backends/webgl/ops/conv.ts
```diff
@@ -153,19 +146,19 @@ export class WebGLUnpackedConv extends Conv {
     const programManager = inferenceHandler.session.programManager;
     if (!this.artifacts) {
       this.artifacts = [];
-      const programInfos = this.createProgramInfos(inferenceHandler, inputs);
+      const programInfos = this.createProgramInfo(inferenceHandler, inputs);
```
xzhu1900, line 149: >createProgramInfo [](start = 32, length = 17)

To Avoid the confusion between this special createProgramInfo with `WebGLOperator`'s abstract frunction, maybe we can rename it to `createProgramInfoArray`?

lib/backends/webgl/ops/conv.ts
```diff
@@ -153,19 +146,19 @@ export class WebGLUnpackedConv extends Conv {
...
       for (let i = 0; i < programInfos.length; ++i) {
         const artifact = inferenceHandler.session.programManager.build(programInfos[i]);
         this.artifacts.push(artifact);
       }
     }
-    const runDatas = this.createRunDatas(inferenceHandler, this.artifacts.map(a => a.programInfo), inputs);
+    const runDatas = this.createRunData(inferenceHandler, this.artifacts.map(a => a.programInfo), inputs);
```
xzhu1900, line 155: >createRunData [](start = 26, length = 13)

Similar to `createProgramInfo`, what do you think of `createRunDataArray`? And also update the return value's name from `runDatas` to `runDataArray`?

lib/backends/webgl/ops/conv-pack.ts
```diff
@@ -0,0 +1,97 @@
...
+    programManager.run(this.artifacts[2], runDataMatmul);
+    const matmulOutput = runDataMatmul.outputTextureData.tensor;
+
+    // reshape output
+    const outputShapeTensor = new Tensor([outputShape.length], 'int32');
+
+    for (let i = 0; i < outputShape.length; i++) {
+      outputShapeTensor.data[i] = outputShape[i];
+    }
+    assert(this.artifacts.length > 1, () => 'expect at least 3 artifacts created');
```
xzhu1900, line 86: >> 1 [](start = 33, length = 3)

>2?

lib/backends/webgl/ops/conv.ts
duli2012: in PR 289.

lib/backends/webgl/ops/conv-pack.ts
duli2012: in PR 289.

PR #279, opened 2021-03-30
Ort Format Support

liuziyue:
Support ORT format models. 
- Load Ort format models using "ort_format_schema.ts" compiled from [ort.fbs](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/flatbuffers/schema/ort.fbs) with the flatbuffer flatc tool. 
- Generate graph accordingly
- mnist.ort as a test case


lib/attribute.ts
```diff
@@ -3,6 +3,7 @@
 
 import Long from 'long';
 import {onnx} from 'onnx-proto';
+import {onnxruntime} from './ort_format_schema';
```
hanbitmyths, line 6: is it possible to change a exported name other than onnxruntime or to set an alias for onnxruntime.experimental.fbs?

lib/attribute.ts
```diff
@@ -27,13 +28,16 @@ type ValueTypes = Attribute.DataTypeMap[Attribute.DataType];
...
 
 export class Attribute {
-  constructor(attributes: onnx.IAttributeProto[]|null|undefined) {
+  constructor(attributes: onnx.IAttributeProto[]|null|undefined|onnxruntime.experimental.fbs.Attribute[]) {
     this._attributes = new Map();
     if (attributes !== null && attributes !== undefined) {
       for (const attr of attributes) {
-        this._attributes.set(attr.name!, [Attribute.getValue(attr), Attribute.getType(attr)]);
+        if (attr instanceof onnx.AttributeProto) {
+          this._attributes.set(attr.name, [Attribute.getValue(attr), Attribute.getType(attr)]);
```
hanbitmyths, line 36: null check is gone here. is it not required?

lib/attribute.ts
```diff
@@ -27,13 +28,16 @@ type ValueTypes = Attribute.DataTypeMap[Attribute.DataType];
 type Value = [ValueTypes, Attribute.DataType];
 
 export class Attribute {
-  constructor(attributes: onnx.IAttributeProto[]|null|undefined) {
+  constructor(attributes: onnx.IAttributeProto[]|null|undefined|onnxruntime.experimental.fbs.Attribute[]) {
```
hanbitmyths, line 31: nit: move null|undefined to the last

lib/attribute.ts
```diff
@@ -93,45 +97,50 @@ export class Attribute {
...
+  private static getValue(attr: onnx.IAttributeProto|onnxruntime.experimental.fbs.Attribute) {
+    const attrType =
+        attr instanceof onnx.AttributeProto ? attr.type : (attr as onnxruntime.experimental.fbs.Attribute).type();
+    if (attrType === onnx.AttributeProto.AttributeType.GRAPH || attrType === onnx.AttributeProto.AttributeType.GRAPHS) {
       throw new Error('graph attribute is not supported yet');
     }
 
-    const value = this.getValueNoCheck(attr);
+    const value = attr instanceof onnx.AttributeProto ?
+        this.getValueNoCheck(attr) :
```
hanbitmyths, line 133: isn't it better to make getValueNoCheck handle both onnx and ort?

lib/attribute.ts
```diff
@@ -143,27 +152,39 @@ export class Attribute {
...
+      }
     }
 
     // cast Uint8Array to string
-    if (attr.type === onnx.AttributeProto.AttributeType.STRING) {
-      const utf8String = value as Uint8Array;
-      return Buffer.from(utf8String.buffer, utf8String.byteOffset, utf8String.byteLength).toString();
+    if (attrType === onnx.AttributeProto.AttributeType.STRING) {
+      // string in onnx attribute is of uint8array type
+      if (attr instanceof onnx.AttributeProto) {
```
hanbitmyths, line 174: leave a comment about why string and strings in ort is not here?

lib/graph.ts
```diff
@@ -102,12 +104,19 @@ class Value implements Graph.Value {
...
-  constructor(_nodeProto: onnx.INodeProto) {
-    this.name = _nodeProto.name!;
-    this.opType = _nodeProto.opType!;
+  constructor(_nodeProto: onnx.INodeProto|onnxruntime.experimental.fbs.Node, name?: string) {
+    if (_nodeProto instanceof onnx.NodeProto) {
+      this.name = _nodeProto.name;
+      this.opType = _nodeProto.opType;
+      this.attributes = new Attribute(_nodeProto.attribute);
+    } else if (_nodeProto instanceof onnxruntime.experimental.fbs.Node) {
+      this.name = name ? name : _nodeProto.name()!;
```
hanbitmyths, line 113: name ?? _nodeProto.name()

lib/graph.ts
```diff
@@ -130,13 +139,19 @@ class GraphImpl implements Graph, Graph.Transformer {
...
-  constructor(graph: onnx.IGraphProto, graphInitializer?: Graph.Initializer) {
+  constructor(graph: onnx.IGraphProto|onnxruntime.experimental.fbs.Graph, graphInitializer?: Graph.Initializer) {
     if (!graph) {
       throw new TypeError('graph is empty');
     }
 
     // build the graph - will throw exceptions if something fatal is detected
-    this.buildGraph(graph);
+    if (graph instanceof onnx.GraphProto) {
+      this.buildGraph(graph);
```
hanbitmyths, line 149: make buildGraph handle both onnx and ort, so we have buildGraph, buildGraphFromOnnxFormat, buildGraphFromOrtFormat?

lib/graph.ts
```diff
@@ -321,6 +336,173 @@ class GraphImpl implements Graph, Graph.Transformer {
...
+          inputValueNames.push(inputName);
+        }
+      }
+    }
+    // check initializers
+    for (let i = 0; i < graph.initializersLength(); i++) {
+      const initializer = graph.initializers(i)!;
+      let index = dataIndices.get(initializer.name()!);
+      const dims = ProtoUtil.tensorDimsFromORTFormat(initializer);
+      const type = ProtoUtil.tensorDataTypeFromProto(initializer.dataType());
```
hanbitmyths, line 387: put dims and type into below if statement

lib/model.ts
```diff
@@ -1,10 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import {flatbuffers} from 'flatbuffers';
 import {onnx} from 'onnx-proto';
 
 import {Graph} from './graph';
 import {OpSet} from './opset';
+import {onnxruntime} from './ort_format_schema';
```
hanbitmyths, line 9: how to update ort format schema when new version is available?

liuziyue: Currently there's no automated way to update it. In onnxruntime, this file is also manually updated with a command. [ORT File Format](https://github.com/microsoft/onnxruntime/tree/d42399e1b07ce61e95aae88bc6b6ea5dcaae2011/onnxruntime/core/flatbuffers/schema#ort-file-format). We'll have to keep track of the version history from onnxruntime, and update the schema accordingly. Maybe we can provide a readme with the update instructions?

lib/attribute.ts
liuziyue: It's not required as we have checked that `attr` is an instance of `onnx.AttributeProto`.

lib/model.ts
hanbitmyths: We may provide a readme and also make a note at ort document.

hanbitmyths (commented):
Please, fix build breaks.

hanbitmyths (approved):

PR #283, opened 2021-04-02
add packed reshape kernel

xzhu1900:
This PR includes the implementation of packed reshape kernel. Corresponding test is included in unittest and op test.

It also includes some fix for texture coordinate calculation and unpack cache bug.


duli2012 (approved):

lib/backends/webgl/ops/reshape-packed.ts
```diff
@@ -0,0 +1,174 @@
...
+import {TextureLayout} from '../types';
+
+import {unpackFromChannel} from './packing_utils';
+
+export class WebGLReshapePacked extends Reshape implements WebGLOperator {
+  run(inferenceHandler: WebGLInferenceHandler, inputs: Tensor[]): Tensor[] {
+    return inferenceHandler.run(this, inputs);
+  }
+  createProgramInfo(handler: WebGLInferenceHandler, inputs: Tensor[]): ProgramInfo {
+    if (inputs.length !== 2) {
```
duli2012, line 19: Resize -> Reshape

xzhu1900: Just saw this comment.. I will correct it in the next PR.

PR #284, opened 2021-04-06
enable unpacked conv to work in mixed texture layout inferencing

xzhu1900:
This PR adds logic to enable unpacked conv op to work in an inference session where both packed kernel and unpacked kernel are used. Now we can inference the teams model with all supported packed kernel with correct result. 

This PR also fixed a kernel cache bug.


duli2012 (approved):

xzhu1900:
reopen to trigger CI.

PR #285, opened 2021-04-09
WebGL profiling tool

duli2012:
Adding profiling tool to generate kernel level execution time for WebGL backend.


lib/backends/webgl/webgl-context.ts
```diff
@@ -464,4 +468,64 @@ export class WebGLContext {
...
+    if (this.version === 2) {
+      const gl2 = this.gl as WebGL2RenderingContext;
+      const ext = this.disjointTimerQueryWebgl2Extension;
+
+      const query = gl2.createQuery() as WebGLQuery;
+      gl2.beginQuery(ext.TIME_ELAPSED_EXT, query);
+      return query;
+    } else {
+      // TODO: add webgl 1 handling.
+      throw new Error(`Not support WebGL 2`);
```
xzhu1900, line 482: >`Not support WebGL 2` [](start = 22, length = 21)

Maybe we can rephrase this error message to be more accurate, as this is not about WebGL2. How about "WebGL1 profiling currently not supported"?

Same for the error messages below.

lib/backends/webgl/webgl-context.ts
```diff
@@ -464,4 +468,64 @@ export class WebGLContext {
...
+  }
+
+  isTimerResultAvailable(query: WebGLQuery): boolean {
+    let available = false, disjoint = false;
+    if (this.version === 2) {
+      const gl2 = this.gl as WebGL2RenderingContext;
+      const ext = this.disjointTimerQueryWebgl2Extension;
+
+      available = gl2.getQueryParameter(query, gl2.QUERY_RESULT_AVAILABLE);
+      disjoint = gl2.getParameter(ext.GPU_DISJOINT_EXT);
```
xzhu1900, line 505: Curious in what scenario would available be false?

docs/development.md
```diff
@@ -56,3 +56,12 @@ To debug on an Android device, please refer to the following steps:
...
 - follow the steps above to debug.
+
+### Performance profiling
+To profile the kernel level execution time, please run the following command:
+npm test -- model <model_name> --profile
+
+It generates raw perf data for each kernel, you may want use tools/parse-profiler.ts to parse it, e.g.:
+
+npm test -- model resnet50 -b=webgl --perf --profile > resnet50_perf.txt
+node tools/parse-profiler.js < resnet50_perf.txt > parsed_resnet50_perf.txt
```
xzhu1900, line 67: Looks like CI is failing complaining formatting this file.

lib/backends/webgl/webgl-context.ts
duli2012: The result  is not available until the profiled kernel execution completes.

docs/development.md
duli2012: fixed.

xzhu1900 (approved):

PR #290, opened 2021-04-27
add texture cache support for pack/unpack mode

xzhu1900:
This change includes the texture cache support from pack/unpack mode. It also contains a handful other bug fixes.


xzhu1900:
The change is squashed from an old branch xuzhu/teams_eval_working2, which contains various commits, other branches merge since January, and hard to isolate out for a single PR.

lib/tensor.ts
```diff
@@ -31,16 +33,13 @@ export declare namespace Tensor {
   export type FloatType = Tensor.DataTypeMap['float32']|Tensor.DataTypeMap['float64'];
   export type NumberType = BooleanType|IntegerType|FloatType;
 
-  export interface Id {
-    // this field helps typescript to perform type check, comparing to use `Id` as an alias of object.
-    _tensorDataId_unused?: never;
-  }
+  export type Id = number;
```
xzhu1900, line 36: personally I feel a unique number is better as the ID than the tensor object. I feel it is much easier for me to debug the texture cache with a number as the key. 

This is more like my preliminary modification. Many things like multi-threading would affect this implementation. I think we discussed about adopting a more debug-friendly type for the ID. Depending on the scope of work, I'll either add it here, or open a new PR for this.

lib/backends/webgl/glsl-coordinate-lib.ts
```diff
@@ -1190,6 +1190,14 @@ export class CoordsGlslLib extends GlslLib {
       result[funcName] = new GlslLibRoutine(
           this.getValueFromSingle(name, rank, layout.width, layout.height, true),
           [`shapeUtils.indicesToOffset${funcName}`, `coordinates.offsetToCoords`, `fragcolor.getColorAsFloat`]);
+      funcName = `_${name}_Pack`;
```
duli2012, line 1193: Just a reminder, we removed this change in conv_pack PR.

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -47,6 +49,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['Clip', '', '6-10', () => new WebGLClip()],
   ['Concat', '', '4+', () => new WebGLConcat()],
   ['Conv', '', '1+', () => new WebGLConv()],
+  ['ConvPacked', '', '1+', () => new WebGLConvPacked()],
```
duli2012, line 52: Same as above. just be careful when you check in.

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -68,6 +71,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['Less', '', '7+', () => new binaryOps.WebGLBinaryOp(NUMBER_TYPES, binaryOps.glslLess(), undefined, 'bool')],
   ['Log', '', '6+', () => new unaryOps.WebGLUnaryOp(FLOAT_TYPES, unaryOps.glslLog())],
   ['MatMul', '', '1+', () => new WebGLMatMul()],
+  ['MatMulPacked', '', '1+', () => new WebGLMatMulPacked()],
```
duli2012, line 74: another one.

lib/backends/webgl/op-resolve-rules.ts
```diff
@@ -84,7 +88,7 @@ export const WEBGL_OP_RESOLVE_RULES: ReadonlyArray<OpSet.ResolveRule> = [
   ['ReduceSum', '', '1+', () => new reduceOps.WebGLReduceSum()],
   ['ReduceSumSquare', '', '1+', () => new reduceOps.WebGLReduceSumSquare()],
   ['Relu', '', '6+', () => new unaryOps.WebGLUnaryOp(FLOAT_TYPES, unaryOps.glslRelu())],
-  ['Reshape', '', '5+', () => new WebGLReshape()],
+  ['Reshape', '', '5+', () => new WebGLReshapePacked()],
```
duli2012, line 91: WebGLReshapePacked() => WebGLReshape()

lib/tensor.ts
```diff
@@ -169,7 +168,8 @@ export class Tensor {
       /**
        * get the data ID that used to map to a tensor data
        */
-      public readonly dataId: Tensor.Id = {}) {
+      public readonly dataId: number = globalId) {
+    globalId += 1;
```
duli2012, line 172: do we need synchronization here?

lib/tensor.ts
Tixxx: +1 for this change, using an empty object as a key provides no traceability. Consider using a UUID or GUID generator library for uniqueness and thread safety(althoguh i dont think that would be necessary since browser tabs are mainly single-threaded).

lib/backends/webgl/glsl-coordinate-lib.ts
xzhu1900: Thanks. Reverted.

lib/backends/webgl/op-resolve-rules.ts
xzhu1900: Done

lib/backends/webgl/op-resolve-rules.ts
xzhu1900: done. Thanks.

lib/tensor.ts
xzhu1900: The measurement shows the perf diff between using number and UUID string is negligible. Updated using UUID. Thanks.

lib/tensor.ts
xzhu1900: not if browser tab is using single thread -- and it seems it is the case.

duli2012 (approved):

PR #291, opened 2021-04-27
Support super resolution models and benchmarking (WIP)

Tixxx:
This PR adds the following support:
depth to space kernel to run office super resolution model
6d tensor support in coordinate library to run model zoo super resolution
a super resolution benchmarking tool that compares the performance of onnxjs to the latest tfjs with profiling info.

This pr also exposes a logger config(only contains a log level filed) to the inference session so that logger's log level can be configured from onnxjs api.

I mark this as WIP for now since some tests are still failing.


benchmark/super_resolution_model_zoo/README.md
```diff
@@ -0,0 +1,38 @@
+# Benchmarks
```
duli2012, line 1: could we generalize this sub project to other models other than SR, which would be very useful feature?

benchmark/super_resolution_model_zoo/LICENSE
```diff
@@ -0,0 +1,5 @@
+# Licenses
```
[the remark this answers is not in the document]
Tixxx: Since we put tfjs in our benchmark, we need to specify tfjs's license here to state that we are using some other project. I think i included a README in the same folder.

benchmark/super_resolution_model_zoo/README.md
Tixxx: Yes i put a TODO in the main index.js file to improve this benchmark to take arbitrary models and configs. For that, we would need a benchmark driver similar to our test-runner-cli. I will do a followup change just for the benchmarking tool.

PR #293, opened 2021-05-03
swap last 2 dims in packed concat for reversed WH

Tixxx:
swap last 2 dims in packed concat for reversed WH

### DIFF
--- a/lib/backends/webgl/glsl-coordinate-lib.ts
+++ b/lib/backends/webgl/glsl-coordinate-lib.ts
@@ -89,7 +89,7 @@ export class CoordsGlslLib extends GlslLib {
    * Generates code for packed output sampler.
    */
   protected getPackedOutputSamplingSnippet(outputLayout: TextureLayout): {[name: string]: GlslLibRoutine;} {
-    const outShape = outputLayout.shape;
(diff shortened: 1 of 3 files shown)


lib/backends/webgl/glsl-coordinate-lib.ts
```diff
@@ -264,8 +264,9 @@ export class CoordsGlslLib extends GlslLib {
    */
   protected getOutputPacked3DCoords(shape: [number, number, number], texShape: [number, number]): GlslLibRoutine {
     const packedTexShape = [texShape[0], texShape[1]];
-    const texelsInLogicalRow = shape[2];
-    const texelsInBatch = texelsInLogicalRow * shape[1];
+    const texelsInLogicalRow = Math.ceil(shape[2] / 2);
```
xzhu1900, line 267: After so many tests, I thought we had a pretty good coverage on this library... :0

xzhu1900 (approved):

lib/backends/webgl/glsl-coordinate-lib.ts
Tixxx: I know.. this was really subtle, only triggered by some specific tensor/tex shapes.

PR #294, opened 2021-05-04
Fix an error caused by glsl mod() function

duli2012:


xzhu1900 (approved):
Signed off. Thanks for fixing it!

One quick question: what's the plan for this repo? Shall we also cherry-pick it to ORT?

lib/backends/webgl/ops/reshape-packed.ts
```diff
@@ -115,8 +119,26 @@ export class WebGLReshapePacked extends Reshape implements WebGLOperator {
     };
   }
   createRunData(handler: WebGLInferenceHandler, programInfo: ProgramInfo, inputs: Tensor[]): RunData {
-    const inputTDs =
-        [handler.getOrCreateTextureData(inputs[0], handler.getOrCreateTextureLayout(inputs[0], 1, false, [], false))];
+    let inputTDs: [TextureData];
+    const originalInputLayout = handler.getOrCreateTextureLayout(inputs[0], 1, false, [], false);
+    const originalInputTD = handler.getOrCreateTextureData(inputs[0], originalInputLayout, false);
+
+    if (this.needSqueezeInputData) {
```
xzhu1900, line 126: >    if (this.needSqueezeInputData) [](http://example.com/codeflow?start=0&length=34)

I vaguely remember this value could be undefined for following runs when the instance is retrieved from cache thus createProgramInfo() is skipped. I need to verify it though. If that's the case, then this statement could be falsely skipped and cause some correctnesss issue.

duli2012: fixed

